PR #32, opened 2020-11-23
Feature/createproject

denise-khuu:
createProject() was added and changeCurrentDir was used in previous functions when needed. 
Still need to add dynamically changing cd template, based on current dir.


SchettlerKoehler (dismissed):

runners/console/index.ts
```diff
@@ -62,6 +66,10 @@ export class Console extends Runner {
         console.log("assertCobiGenJava");
     }
 
+    async assertCreateProject(step: Step, command: Command, result: RunResult) {
+        console.log("assertCobiGenJava");
```
SchettlerKoehler, line 70: Change to assertCreateProject

runners/katacoda/templates/cd.md
```diff
@@ -0,0 +1,4 @@
+
+We have to start at the right folder
```
SchettlerKoehler, line 2: Better something like: Please change the folder to <% dir: %>

runners/katacoda/index.ts
```diff
@@ -20,7 +20,8 @@ export class Katacoda extends Runner {
     private setupScripts: KatacodaSetupScript[] = [];
     private assetManager: KatacodaAssetManager;
     private setupDir: string;
-
+    private currentDir: string = "root";
```
SchettlerKoehler, line 23: /root

runners/katacoda/index.ts
```diff
@@ -66,6 +67,8 @@ export class Katacoda extends Runner {
     }
 
     runInstallDevonfwIde(step: Step, command: Command): RunResult {
+        let cdCommand = this.changeCurrentDir("root");
```
SchettlerKoehler, line 70: /root

runners/katacoda/index.ts
```diff
@@ -110,6 +117,24 @@ export class Katacoda extends Runner {
         });
         this.renderTemplate("cobiGenJava.md", this.outputPathTutorial + "step" + (this.stepsCount++) + ".md", { text: step.text, textAfter: step.textAfter, javaFile: params[0], cobiGenTemplates: cobiGenTemplates });
         return null;
+
+    }
+
+    runCreateProject(step: Step, command:Command): RunResult {
+
+        // generate template to change directory, if the current directory is not equal to the required start directory
+       let cdCommand = this.changeCurrentDir(path.join("root", "devonfw"));
```
SchettlerKoehler, line 126: /root

runners/katacoda/templates/createProject.md
```diff
@@ -0,0 +1,16 @@
...
+<%= cdCommand; %>
+
+Navigate to the 'workspaces/main/' folder in your devonfw installation directory.
+`cd workspaces/main/`{{execute}}
+
+Now you can use devon to setup a <%= language; %>  project for you by executing the following command.
+`devon <%= language; %> create com.example.application.<%= name; %>`{{execute}}
+
+Switch into the newly create project directory.
+`cd cobigenexample/`{{execute}}
```
SchettlerKoehler, line 14: The folder name is the name of the project

runners/katacoda/index.ts
```diff
@@ -110,6 +117,28 @@ export class Katacoda extends Runner {
...
+        // generate template to change directory, if the current directory is not equal to the required start directory
+       let cdCommand = this.changeCurrentDir(path.join("/root", "devonfw"));
+
+       this.steps.push({
+           "title": "Create a new project",
+           "text": "step" + this.stepsCount + ".md"
+       });
+
+        //update current directory
+       this.currentDir = path.join(this.currentDir, "workspace", "main","cobigenexample");
```
SchettlerKoehler, line 138: Remember to change this when you change the template

localBuildRun.sh
```diff
@@ -3,4 +3,5 @@ cp engine/parser.def build/engine/parser.def
 cp -r ../tutorials build/playbooks
 cp -r environments build
 cp -r runners build
+npm test
 node build/engine/run.js
```
GuentherJulian, line 7: In localBuildRun.ps1 you throw an error if a test fails. This should be consistent in both files.

localBuildTest.ps1
```diff
@@ -0,0 +1,7 @@
+tsc
+Copy-Item -Force -Recurse -Path $PSScriptRoot\engine\parser.def -Destination $PSScriptRoot\build\engine\parser.def
+Copy-Item -Force -Recurse -Path $PSScriptRoot\..\tutorials -Destination $PSScriptRoot\build\playbooks
+Copy-Item -Force -Recurse -Path $PSScriptRoot\environments\ -Destination $PSScriptRoot\build
+Copy-Item -Force -Recurse -Path $PSScriptRoot\runners\ -Destination $PSScriptRoot\build
```
GuentherJulian, line 5: Is it necessary to copy all the files in the build folder if only the test cases are to be checked?

localBuildTest.sh
```diff
@@ -0,0 +1,6 @@
+tsc
+cp engine/parser.def build/engine/parser.def
+cp -r ../tutorials build/playbooks
+cp -r environments build
+cp -r runners build
```
GuentherJulian, line 5: Same as in localBuildTest.ps1

runners/katacoda/dirUtils.ts
```diff
@@ -0,0 +1,51 @@
+import { captureRejectionSymbol } from "events";
```
GuentherJulian, line 1: This is never used.

runners/katacoda/dirUtils.ts
```diff
@@ -0,0 +1,51 @@
...
+import * as path from 'path';
+
+export class DirUtils{
+
+    getCdParam(currentDir:string, dir:string):string{
+
```
GuentherJulian, line 7: Better to name 'dir' to 'targetDir' (or something similar) to indicate that this is directory to navigate into.

runners/katacoda/dirUtils.ts
```diff
@@ -0,0 +1,51 @@
...
+        let currentPaths = currentDir.split(path.sep);
+        let dirPaths = dir.split(path.sep);
+        
+        let index;
+        let isEqual = true;
+
+        //returns an empty string, if both variables have the same path 
+        if(currentDir == dir){
+            return "";
+        }
```
GuentherJulian, line 19: This can be checked directly at the top of the method.

runners/katacoda/index.ts
```diff
@@ -110,6 +118,28 @@ export class Katacoda extends Runner {
...
         return null;
+
+    }
+
+    runCreateProject(step: Step, command:Command): RunResult {
+
+        let params = command.parameters; 
+        let language = params[0];
+        let name = params[1]; 
+
```
GuentherJulian, line 129: Here you can use command.parameters[0] and command.parameters[1]  directly.

runners/katacoda/templates/createProject.md
```diff
@@ -0,0 +1,16 @@
+<%= text; %>
+
+## Setting up your java project
+
+<%= cdCommand; %>
+
+Navigate to the 'workspaces/main/' folder in your devonfw installation directory.
+`cd workspaces/main/`{{execute}}
+
+Now you can use devon to setup a <%= language; %>  project for you by executing the following command.
```
maybeec, line 10: either write "use the devon command" or say "use devonfw to setup" as devonfw is a trademark. We need to get the usage clear in our docs

runners/katacoda/templates/createProject.md
```diff
@@ -0,0 +1,16 @@
...
+`devon <%= language; %> create com.example.application.<%= name; %>`{{execute}}
+
+Switch into the newly create project directory.
+`cd <%= name; %>`{{execute}}
```
maybeec, line 14: not sure, whether this is a good idea for maintainability.
We also have a cd step as I have seen. So possibly, we should not walk to different folders when creating a new project

maybeec (changes requested):

maybeec:
@denise-khuu , @GuentherJulian  as of my comments in https://github.com/devonfw-forge/tutorial-compiler/pull/38/files#r534927014 I feel, that we have to create multiple steps for each type of project we want to create.

The reason behind that is quite easy. Here you created a generic createProject step, which a parameter for the language.
The problem with that is, that the project generation will differ tremendously is each stack obviously. So also the katacoda templates etc will describe completely different things. In addition, each project creation might come with a complete different set of parameters to customize the project creation. So there is no "common code" at all I feel, which simply leads to a very complex step implementation with lots of case differentiation.
Finally, the last point here is, that there might be even multiple templates (per architecture) for one programming language. So just differentiating the project creation based on the programming language will not work.

So for me as you already started with it, I would propose to simply remove the first parameter of the step and simply rename the createProject step to createDevon4jProject and then also the katacoda template and so on is fine as it focuses on the steps for that generation only.

runners/katacoda/templates/createProject.md
denise-khuu: Should I also use the cd Template to change to workspace/main or should I keep it that way?

maybeec: yes would be better as maybe you want to create the project at a completely different folder

localBuildRun.sh
```diff
@@ -3,4 +3,8 @@ cp engine/parser.def build/engine/parser.def
 cp -r ../tutorials build/playbooks
 cp -r environments build
 cp -r runners build
+npm test
+if [ $? -eq 1 ]; then 
+    echo 'tests failed' 
+fi
```
GuentherJulian, line 9: In the powershell script you throw an error when tests will fail. This ends the task. Here you just log it to the console and the task continues. Maybe its better to exit here as well to keep it consistent

runners/katacoda/dirUtils.ts
```diff
@@ -0,0 +1,51 @@
...
+                }else{
+                    isEqual = false;
+                    dirPath = path.join(dirPath,'..');
+                }
+            })
+
+            //slice targetPaths to get the relative path
+            targetPaths = targetPaths.slice(index + 1, targetPaths.length);
+        
+            return path.join(dirPath, targetPaths.join(path.sep));
```
GuentherJulian, line 45: When you build the project on your windows machine, then this will generate paths with backslash. In katacoda we need paths with "/". So you have to replace the "\" with "/" before returning the path.

runners/katacoda/index.ts
```diff
@@ -110,6 +118,25 @@ export class Katacoda extends Runner {
...
         this.renderTemplate("cobiGenJava.md", this.outputPathTutorial + "step" + (this.stepsCount++) + ".md", { text: step.text, textAfter: step.textAfter, javaFile: params[0], cobiGenTemplates: cobiGenTemplates });
         return null;
+
+    }
+
+    runCreateDevon4jProject(step: Step, command:Command): RunResult {
+
+        // generate template to change directory, if the current directory is not equal to the required start directory
+       let cdCommand = this.changeCurrentDir(path.join("/root", "devonfw"));
+       let cdTarget = this.changeCurrentDir(path.join(command.parameters[1]));
```
GuentherJulian, line 128: Is it necessary to pass this parameter. In the task for the console runner we pass only the project name (https://github.com/devonfw-forge/tutorial-compiler/commit/a31e110b6652d106552cdd5e9a817100a6b79a28). I think you can assume that the project will always be generated in the workspace/main folder.

maybeec:
Please resolve conflicts

maybeec (approved):

PR #38, opened 2020-11-30
Feature/consoleCreateProject

GuentherJulian:


runners/console/index.ts
```diff
@@ -60,6 +60,20 @@ export class Console extends Runner {
         return result;
     }
 
+    runCreateProject(step: Step, command: Command): RunResult {
+        let result = new RunResult();
+        result.returnCode = 0;
+
```
maybeec, line 66: there is no differentiation on command.parameters[0].
I think we should throw an exception somehow to indicate if a template was used, which does not exist for easier debugging and understanding of failures

documentation/Functions.md
```diff
@@ -1,3 +1,44 @@
...
+
+***
+
+### createProject
+#### parameter 
+1. the programming language that you will use for this project
+2. the project name
+
+#### example 
+createProject("java","cobigenexample")
```
maybeec, line 42: let's not focus on the programming language as a first parameter as there might be different architectures for one programming language bound to different project templates in near future.
So let's focus on a parameter like devon4j to indicate devon4j project template.

runners/console/index.ts
```diff
@@ -60,6 +60,20 @@ export class Console extends Runner {
...
+        let workspaceDir = path.join(this.getWorkingDirectory(), "devonfw", "workspaces", "main")
+        if(this.platform == ConsolePlatform.WINDOWS) {
+            this.executeCommandSync("devon " + command.parameters[0] + " create com.example.application." + command.parameters[1], workspaceDir, result);
```
maybeec, line 69: where is the second parameter here taken into account?

maybeec (changes requested):

documentation/Functions.md
GuentherJulian: Do you mean a differentiation between devon4j, devon4net, etc?
Should we create separate tasks for these?

maybeec: Yes and even within devon4j there might be more than one option to create a project.

I am not sure, possibly it's more simpler to have a task per stack like devon4j vs devon4ng.

documentation/Functions.md
```diff
@@ -35,10 +35,10 @@ cobiGenJava("path/to/java/file/MyEntity.java",[1,3,5,6,8])
 
 ### createProject
 #### parameter 
-1. the programming language that you will use for this project
+1. the template used for project generation
 2. the project name
 
 #### example 
-createProject("java","cobigenexample")
+createProject("devon4j","cobigenexample")
```
maybeec, line 42: I even thought of creating a step for it as there will be no logic at all to be reused. So we should split this into createDevon4jProject("cobigenexample") and further

maybeec (changes requested):

runners/console/index.ts
```diff
@@ -92,6 +107,19 @@ export class Console extends Runner {
         console.log("assertCobiGenJava");
```
maybeec, line 107: let's properly console log that there is no assertion at all. 
We most properly need to address it later

maybeec (changes requested):

maybeec (approved):

PR #40, opened 2020-12-05
Feature/consoleBuildJava

GuentherJulian:


denise-khuu (approved):

runners/console/index.ts
```diff
@@ -103,6 +117,17 @@ export class Console extends Runner {
...
+    async assertBuildJava(step: Step, command: Command, result: RunResult) {
+        let workspaceDir = path.join(this.getWorkingDirectory(), "devonfw", "workspaces", "main");
+
+        new Assertions()
+        .noErrorCode(result)
+        .noException(result)
+        .directoryExits(path.join(workspaceDir, command.parameters[0], "api", "target"))
+        .directoryExits(path.join(workspaceDir, command.parameters[0], "core", "target"))
+        .directoryExits(path.join(workspaceDir, command.parameters[0], "server", "target"));
+    }
```
maybeec, line 129: Can we make sure, that maven succeeded by checking the return code?

maybeec (changes requested):

runners/console/index.ts
GuentherJulian: We check the return code in the first assertion. Or is it possible that maven even fails with a return code of 0?

maybeec: mh good question, let's keep it then this way first. Actually, if the build itself fails, it should always be return code 1.
Thanks for clarification.

maybeec (approved):

PR #46, opened 2020-12-09
Feature/katacodabuild java

denise-khuu:
spelling mistake in cdCommand and currentDir. cdCommand in katacodabuildJava added


runners/katacoda/index.ts
```diff
@@ -154,6 +153,7 @@ export class Katacoda extends Runner {
             "title": "Build the java project",
             "text": "step" + this.stepsCount + ".md"
         });
+
```
GuentherJulian, line 156: The second parameter is optional, so you must first check that the second parameter is set (default is false).
In addition, the second parameter is an indicator of whether the tests should be run. So if the parameter is true, 'maven.test.skip' has to be false. And vice versa.

maybeec (approved):

PR #60, opened 2020-12-14
Feature/katacodabuild java

denise-khuu:


runners/katacoda/index.ts
```diff
@@ -140,15 +140,20 @@ export class Katacoda extends Runner {
...
-        
-        if(command.parameters[1] == true){
-            skipTest = '-Dmaven.test.skip=' + command.parameters[1];
-            skipTestDescr = "We do not need to execute the test cases, so we can skip them by using the option '-Dmaven.test.skip=true'."
+        let skipTest;
+        let skipTestDescr;
+
+        if(command.parameters.length == 2 && command.parameters[1] == true){
+            skipTest = "";
+            skipTestDescr = "";
```
GuentherJulian, line 150: When you initialize the variables before the if statement and check in the condition if the parameter is false, then you can remove the first part of the if statement.

maybeec (approved):

PR #61, opened 2020-12-14
Feature/katacodacreatefile

denise-khuu:


runners/katacoda/templates/createFile.md
```diff
@@ -0,0 +1,14 @@
+<%= text; %>
+
+<%= cdCommand; %>
+
+
+`mkdir -p <%= filePath; %>`{{execute}}
```
GuentherJulian, line 6: Here we should add some text for the user about what this command does.

runners/katacoda/templates/createFile.md
```diff
@@ -0,0 +1,14 @@
...
+
+Switch to the IDE and click 'Copy to Editor' in the following java class. The java file will be created automatically inside the newly created package.
```
GuentherJulian, line 8: The file has not to be a java file in future. So we should keep this text more general. Maybe its better to specify the text in the playbook and pass the description by the text-placeholder into this file.

maybeec (approved):

PR #75, opened 2020-12-22
feature/cloneRepository

TheMarcy:


runners/console/index.ts
```diff
@@ -143,6 +143,19 @@ export class Console extends Runner {
...
     }
 
+    runCloneRepository(step: Step, command: Command): RunResult {
+        let result = new RunResult();
+        result.returnCode = 0;
+
+        let workspaceDir = path.join(this.getWorkingDirectory(), "devonfw", "workspaces", "main");
+        let directorypath = path.join(workspaceDir, command.parameters[0]);
+        
+        this.createFolder(directorypath, false);
```
GuentherJulian, line 153: If you changed the second parameter to 'true', you will probably not get any error messages when building the project and the folder already exists.

runners/console/index.ts
```diff
@@ -222,6 +235,15 @@ export class Console extends Runner {
         .fileContains(filepath, content);
     }
 
+    async assertCloneRepository(step: Step, command: Command, result: RunResult) {
+        let repository = command.parameters[1]
```
denise-khuu, line 239: missing semicolon

runners/console/index.ts
```diff
@@ -222,6 +235,15 @@ export class Console extends Runner {
...
+        let repoName = repository.slice(repository.lastIndexOf("/"), -4);
+        new Assertions()
+        .noErrorCode(result)
+        .noException(result)
+        .directoryExits(path.join(this.getWorkingDirectory(), "devonfw", "workspaces", "main", command.parameters[0], repoName));
+    }
```
GuentherJulian, line 245: Here you can also check if the directory the repository should be cloned into is not empty to ensure that the cloning was successful

runners/katacoda/index.ts
```diff
@@ -209,6 +209,21 @@ export class Katacoda extends Runner {
 
     }
 
+    runCloneRepository(step: Step, command: Command): RunResult {
+
+        let cdCommand = this.changeCurrentDir(path.join("/root", "devonfw", "workspaces", "main"));
+        let workspaceDir = path.join("devonfw", "workspaces", "main");
+        let directoryPath = path.join(workspaceDir, command.parameters[0]).replace(/\\/g, "/");
```
GuentherJulian, line 216: We are not going to use devon ide in every tutorial. The command has to work without ide, as well. So better use the relative path to the root-directory.

GuentherJulian (changes requested):

runners/console/index.ts
denise-khuu: Do we have to change createFile and changeFile, too? They are also written to be solely used in workspaceDir.

GuentherJulian: That's right, then we have to change this too. I think the command should work even without a Devon Ide. @maybeec  what do you think?

runners/katacoda/index.ts
denise-khuu: -Do we have to change createFile and changeFile, too? They are also written to be solely used in workspaceDir.
-That's right, then we have to change this too. I think the command should work even without a Devon Ide. @maybeec what do you think?

denise-khuu (approved):

GuentherJulian (commented):
Please add the documentation for the new feature

runners/katacoda/index.ts
```diff
@@ -228,6 +228,21 @@ export class Katacoda extends Runner {
...
```
GuentherJulian, line 235: You do not need the workspace directory in the directory path for the repo. After the execution of the first cd command the user is already in the workspace directory.

runners/katacoda/templates/cloneRepository.md
```diff
@@ -0,0 +1,15 @@
+<%= text; %>
+
+<%= cdCommand; %>
+
+If the parent directories aren't already in the project, 'mkdir -p' will create them for you. 
+
+`mkdir -p <%= directoryPath; %>`{{execute}}
+
+`cd  <%= directoryPath; %>`{{execute}}
```
GuentherJulian, line 9: Note the case that the first parameter is an empty string. In this case you do not need these two commands

GuentherJulian (changes requested):

runners/console/index.ts
maybeec: Possibly you could also check that git status -s is an empty string. If the string is not empty or the execution returns an error we can be sure that the clone is erroneous.

runners/katacoda/index.ts
```diff
@@ -228,6 +228,26 @@ export class Katacoda extends Runner {
...
+            "title": "Clones Repository " + command.parameters[1],
+            "text": "step" + this.stepsCount + ".md"
+        });
+
+        // checks if path is an empty string
+        if(Boolean(command.parameters[0].trim())) {
+            this.renderTemplate("cloneRepository.md", this.outputPathTutorial + "step" + (this.stepsCount++) + ".md", { text: step.text, textAfter: step.textAfter, cdCommand: cdCommand, directoryPath: directoryPath, repository: command.parameters[1]});
+        } else {
+            this.renderTemplate("cloneRepositoryEmptyString.md", this.outputPathTutorial + "step" + (this.stepsCount++) + ".md", {text: step.text, textAfter: step.textAfter, cdCommand: cdCommand, repository: command.parameters[1]});
+        }
```
GuentherJulian, line 246: The second template is not needed. You can use if statements in the ejs templates. So you can check in the template itself if the given string is not empty and render then some additional text.

GuentherJulian (changes requested):

runners/console/index.ts
TheMarcy: @maybeec To check for an empty String i would need a new Assertion, right?

runners/console/index.ts
```diff
@@ -231,6 +243,28 @@ export class Console extends Runner {
...
+        let repository = command.parameters[1];
+        let repoName = repository.slice(repository.lastIndexOf("/"), -4);
+        let directorypath = path.join(this.getWorkingDirectory(), "devonfw", "workspaces", "main", command.parameters[0], repoName);
+        
+        let process = child_process.spawnSync("git status -s", { shell: true, cwd: directorypath });
+        if(process.status != 0) {
+            result.returnCode = process.status;
+        }
+         // checks if stdout is not empty
+         if(Boolean(process.stdout.toString())) {
```
maybeec, line 256: There is no need to create a boolean object here. Empty strings are falsey

maybeec (changes requested):

runners/console/index.ts
```diff
@@ -231,6 +243,28 @@ export class Console extends Runner {
...
+        
+        let process = child_process.spawnSync("git status -s", { shell: true, cwd: directorypath });
+        if(process.status != 0) {
+            result.returnCode = process.status;
+        }
+
+        if(process.stdout.toString()) {
+            console.log("Current git status: " + process.stdout.toString());
+            result.returnCode = 1;
+        }
```
GuentherJulian, line 259: Can we move this part in an own assertion? Or is this too special and only needed for this use case? What do you think @maybeec?
At least we should throw an exception with a custom error message. If we just set the return code to 1, the assertion noErrorCode will throw the error. And this assertion just log that the return code is not 0. So we have no information about the information we get from the git status command.

GuentherJulian (changes requested):

runners/console/index.ts
maybeec: You are right. Would be better to provide a proper assertion for that. Something like assertRepositoryToBeClean

runners/katacoda/templates/cloneRepository.md
```diff
@@ -0,0 +1,17 @@
+<%= text; %>
+
+<%= cdCommand; %>
+
+<% if (string) { %>
+If the parent directories aren't already in the project, 'mkdir -p' will create them for you.
```
GuentherJulian, line 6: Can't we simply check if directoryPath is not empty? Then we don't need this extra variable.

denise-khuu: Can't we just use 'git clone <repository> <path>' to create the directories? I believe that it will also generate the missing directories.

denise-khuu: git clone [repo] [path]

GuentherJulian: Yes, that will work too. Maybe it's better this way. Then the user has to change the directory only once.

GuentherJulian (approved):

maybeec (approved):

PR #76, opened 2020-12-22
WIP: feature/consoleDockerCompose

MarcelDiessner:


TheMarcy:
I don't really know what to check in the assertDockerCompose

MarcelDiessner:
@maybeec Only the linux pipeline succeeds, but windows always fails, i tried multiple projects
. Locally on my windows pc it works fine. Seems like its a problem with the installed docker. See failed jobs for logs.

GuentherJulian:
@MarcelDiessner, @maybeec I think for most projects the command will not work.
I can reproduce the error thrown in the build for the my-thai-star project when switching to windows containers in docker. In the project, 'bridge' is set as the network driver for docker. This driver is not supported on windows. When switching back to linux containers, the docker-compose command works as expected.
The problem is that linux containers are not supported on Windows Server 2019, as explained here: https://github.com/actions/virtual-environments/issues/1143#issuecomment-651178811

MarcelDiessner:
I think with my latest changes i should be able to make it run on windows pipeline, but the path of the DockerCli.exe is wrong. Is there a way to get the path of this file.

GuentherJulian:
> I think with my latest changes i should be able to make it run on windows pipeline, but the path of the DockerCli.exe is wrong. Is there a way to get the path of this file.

Maybe you can extract the docker installation directory from the environment variables.

documentation/Functions.md
```diff
@@ -152,6 +153,19 @@ npmInstall("my-thai-star/angular")
...
+### dockerCompose
+#### parameter 
+1. Path to the directory where the docker-compose.yml file is located, relative to workspace.
+2. Assertion information. Only needed for the console runner to check if the server was started properly.
+#### example 
+dockerCompose("my-thai-star", { "startupTime": 600, "port": 8081 })
+
+##### Assertion information
+startupTime = Time in seconds to wait before checking if the server is running
+port: Port on which the server is running
```
GuentherJulian, line 165: Will this work for every project? I think in some cases we have to specify a path, too?

engine/runner.ts
```diff
@@ -110,4 +110,22 @@ export abstract class Runner {
...
     }
+
+    commandIsSkippable(command: String): Boolean {
+        let returnVal = false; 
+        let runner = this.getVariable("skipCommands." + this.getRunnerName());
+        if(runner) {
+            if(runner instanceof Array) {
+                for (var i=0; i < runner.length; i++){
+                    if (runner[i] == command) {
+                        returnVal = true;
```
GuentherJulian, line 121: Here you can use runner.indexOf(command) instead of a loop.

runners/console/index.ts
```diff
@@ -9,6 +9,7 @@ import * as path from 'path';
 import * as child_process from "child_process";
 import * as fs from "fs";
 import * as psList from "ps-list";
+import { info } from "console";
```
GuentherJulian, line 12: Not used

runners/console/index.ts
```diff
@@ -207,6 +209,35 @@ export class Console extends Runner {
...
+
+        let process = this.executeCommandAsync("docker-compose up", filepath, result);
+        // process.stderr.setEncoding('utf-8');
+        // process.stderr.on('data', (data) => {
+        //    console.log("stderr: " + data);
+        // });
+        // process.stdout.setEncoding('utf8');
+        // process.stdout.on('data', (data) => {
+        //     console.log(data);
+        // });
```
GuentherJulian, line 227: Remove the comments

GuentherJulian (changes requested):

documentation/Functions.md
MarcelDiessner: a path to what exactly?

GuentherJulian, line 165: You have forgotten the path in the assertion information

GuentherJulian (changes requested):

GuentherJulian (approved):

Issue #50, opened 2020-12-10
katacoda runner: buildNg


GuentherJulian:
Implement the command buildNg in the katacoda runner. Gets the path to the angular project as parameter.

PR #78, opened 2020-12-23
WIP: feature/katacodaBuildNg

MarcelDiessner:
#50

### DIFF
--- a/runners/katacoda/index.ts
+++ b/runners/katacoda/index.ts
@@ -210,17 +210,14 @@ export class Katacoda extends Runner {
     }
 
     runBuildNg(step: Step, command: Command): RunResult {
-        let cdCommand = this.changeCurrentDir(path.join("/root", "devonfw", "workspaces", "main"));
-
-        let projectPath = path.join(command.parameters[0].substring(0,path.join(command.parameters[0]).lastIndexOf(path.sep))).replace(/\\/g, "/");
-        let projectName = path.basename(path.join(command.parameters[0]));
+        let cdCommand = this.changeCurrentDir(path.join("/root", "devonfw", "workspaces", "main", command.parameters[0]));
 
         this.steps.push({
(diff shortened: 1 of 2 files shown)


GuentherJulian:
@TheMarcy dont't forget to modify the function summary in the documentation folder when adding a new feature.

Please add also a branch with the same name as the feature branch to your fork of the tutorials repo, to force the build pipeline to use your command. At the moment the build pipeline uses the main branch of your fork where the new command is not implemented.

runners/katacoda/index.ts
```diff
@@ -240,6 +240,20 @@ export class Katacoda extends Runner {
...
+
+    runBuildNg(step: Step, command: Command): RunResult {
+        let cdCommand = this.changeCurrentDir(path.join(this.getVariable(this.workspaceDirectory), command.parameters[0]));
+
+        this.steps.push({
+            "title": "Build the Angular Project",
+            "text": "step" + this.stepsCount + ".md"
+        });
+
+        this.renderTemplate("buildNg.md", this.outputPathTutorial + "step" + (this.stepsCount++) + ".md", { text: step.text, textAfter: step.textAfter, cdCommand: cdCommand, outputDir: command.parameters[1], useDevonCommand: this.getVariable(this.useDevonCommand) });
```
GuentherJulian, line 252: Wouldn't this cause an error if no second parameter is specified for the build folder?

GuentherJulian (changes requested):

runners/katacoda/index.ts
MarcelDiessner: If you look at the tutorial, i tested both cases and it worked as it should. If there is no value command.parameters[1] ten buildDir is null and in the markdown template it checks if the buildDir is not null.

GuentherJulian (approved):

Issue #51, opened 2020-12-10
console runner: buildNg


GuentherJulian:
Implement the command buildNg in the console runner. Gets the path to the angular project as parameter.

PR #81, opened 2020-12-23
WIP: feature/consoleBuildNg

MarcelDiessner:
#51 
Needs cloneRepository for successful testing

### DIFF
--- a/documentation/Functions.md
+++ b/documentation/Functions.md
@@ -96,3 +96,11 @@ A placeholder is optional. If you do not define a placeholder, the content in th
 Please try not to use custom placeholders. Keep in mind that you might want to build the project before changing them. Custom placeholders with a comment-syntax (e.g. "//PLACEHOLDER") will be removed by the console-environment and others might cause errors.
 
 ***
+
+### buildNg
+#### parameter 
+1. Path to the angular project relative to workspace
+#### example 
+buildNg("exampleAngularProject")
+
+***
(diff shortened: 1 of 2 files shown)


TheMarcy:
Currently buildNg will always compile in the folder dist, because of the usage of 'ng build --output-path dist' and the will ignore the outputPath in angular.json, because assertion will always look for dist. Alternatively, i would need to get the outputPath key from angular.json to control successful compilation. Which way would be better? @GuentherJulian

GuentherJulian16:
> Currently buildNg will always compile to the folder dist, because of the usage of 'ng build --output-path dist' and will ignore the outputPath in angular.json, because the assertion will always look for dist directory. Alternatively, i would need to get the outputPath key from angular.json to control successful compilation. Which way would be better? @GuentherJulian

We can pass the output folder as a second optional parameter. If the user pass the paramter to the function we can simply check in the assertion if this directory exists. If no output parameter is given then we use the normal output path specified in the angular.json file. Is there a simple way to extract this path from the file?

GuentherJulian:
> Currently buildNg will always compile to the folder dist, because of the usage of 'ng build --output-path dist' and will ignore the outputPath in angular.json, because the assertion will always look for dist directory. Alternatively, i would need to get the outputPath key from angular.json to control successful compilation. Which way would be better? @GuentherJulian

We can pass the output folder as a second optional parameter. If the user pass the paramter to the function we can simply check in the assertion if this directory exists. If no output parameter is given then we use the normal output path specified in the angular.json file. Is there a simple way to extract this path from the file?

MarcelDiessner:
Currently I'm not able to find out why the ng build command is executed succesfully in linux pipeline but not on windows.

GuentherJulian (approved):

PR #82, opened 2020-12-23
feature/consolerestoredevonfwide

denise-khuu:


runners/console/index.ts
```diff
@@ -158,6 +166,21 @@ export class Console extends Runner {
         }
     }
 
+    async assertRestoreDevonfwIde(step: Step, command: Command, result: RunResult) {
```
GuentherJulian, line 169: Can't we simply use the assert method of installDevonfwIde method here?

documentation/Functions.md
```diff
@@ -18,6 +18,17 @@ installDevonfwIde(["java","mvn"], "2020.08.001")
...
 ***
 
+### restoreDevonfwIde
+#### parameter
+1. The tools you want to install within the devonfw ide: string array
+2. Optional: The version of the ide to install
+#### example
+restoreDevonfwIde(["java","mvn"], "2020.08.001")
+#### details 
+The installation of the devonfw IDE is executed in a startup script.
```
GuentherJulian, line 28: We should also mention that this is only done in the case of katacoda

GuentherJulian (changes requested):

runners/console/index.ts
```diff
@@ -52,6 +52,14 @@ export class Console extends Runner {
         return result;
     }
 
+
+    runRestoreDevonfwIde(step: Step, command: Command): RunResult {
+        let result = new RunResult();
+        result.returnCode = 0;    
+        result = this.runInstallDevonfwIde(step, command);
```
denise-khuu, line 59: Are those lines necessary? Or is 
let result = this.runInstallDevonfwIde(step, command) 
also possible?

GuentherJulian: These are not necessary. We can simply write return this.runInstallDevonfwIde(step, command)

GuentherJulian (approved):

PR #86, opened 2021-01-06
feature/katacodaRunServerJava

denise-khuu:


runners/katacoda/templates/buildJava.md
```diff
@@ -4,7 +4,7 @@
  
 Use the following devon command to build the java project.
```
GuentherJulian, line 6: It is no devon command, when you are not using 'devon' within the command

runners/katacoda/templates/runServerJava.md
```diff
@@ -0,0 +1,7 @@
+<%= cdCommand; %>
+
+Start the server in terminal <%= terminalId; %> by running the maven command 'mvn spring-boot:run'.
+
+`mvn spring-boot:run`{{execute T<%= terminalId; %> <% if (interrupt) { %>interrupt<% } %>}}
+
+<%= textAfter; %>
```
GuentherJulian, line 7: Here we can mention that it may take a few seconds for the server to run. And maybe we can add that the server is stopped and relaunched automatically by executing the command (only if it is running)

GuentherJulian (changes requested):

GuentherJulian (dismissed):

GuentherJulian (approved):

PR #92, opened 2021-01-12
bugfix/katacodaRestoreDevonfwIde

denise-khuu:


runners/katacoda/index.ts
```diff
@@ -111,6 +111,9 @@ export class Katacoda extends Runner {
             "script": "restoreDevonfwIde.sh"
         });
 
+        fs.appendFileSync(path.join(this.getRunnerDirectory(),"templates","scripts", "intro_foreground.sh"), "\n cd devonfw \n . ~/.bashrc \n");
```
GuentherJulian, line 114: You do not have to change the directory first. Just using the '. ~/.bashrc' command is enough.

GuentherJulian (changes requested):

runners/katacoda/index.ts
denise-khuu: Katacoda told me to change to the devonfwIde installation folder:` You are not inside a devonfw-ide installation: /root`

GuentherJulian: Yes, that is correct. But to make the command available in the terminal, it is enough to use the '. ~/.bashrc' command. This will reload the bashrc file in the terminal where the devon command is aliased.

denise-khuu: I just tried to run the  '. ~/.bashrc' command in the "/root"-folder, but couldn't find "DEVON_IDE_HOME" in the environment variables. Are you sure that this is possible?

GuentherJulian (approved):

PR #100, opened 2021-01-26
refactoring/consolePathHandling

denise-khuu:


runners/console/index.ts
```diff
@@ -80,6 +82,9 @@ export class Console extends Runner {
             this.executeCommandSync("bash setup " + path.join(settingsDir, "settings.git").replace(/\\/g, "/"), installDir, result, "yes");
         }
 
+        this.setVariable(this.workspaceDirectory, path.join(this.getWorkingDirectory(), "devonfw", "workspaces", "main"));
```
GuentherJulian, line 85: I think you should use the new way to get the workspace directory in runCreateDevon4jProject, assertCreateDevon4jProject and assertCobiGenJava, too. 
And maybe is it useful to log a warning when a command is used which requires the devon ide, but it's not used.

GuentherJulian (changes requested):

runners/console/index.ts
denise-khuu: Wouldn't it be better to just add the warning without changing the paths. My thoughts were, that the given path is mandatory for the function to work. So it will throw an Error, when the directories don't exist.

GuentherJulian: Then you should revert it in runCobiGenJava. There you handle it the new way.
My thoughts were that if no ide is installed the noErrorCode assertion will throw the error because the devon command is not recognized. So the directories will not be checked at all.

runners/console/index.ts
```diff
@@ -91,28 +96,32 @@ export class Console extends Runner {
     runInstallCobiGen(step: Step, command: Command): RunResult {
         let result = new RunResult();
         result.returnCode = 0;
-
-        this.executeDevonCommandSync("cobigen", path.join(this.getWorkingDirectory(), "devonfw"), result);
+        if(this.getVariable(this.useDevonCommand)){
+            this.executeDevonCommandSync("cobigen", path.join(this.getWorkingDirectory(), "devonfw"), result);
+        }else{
+            console.warn("Devonfw IDE is not installed");
```
GuentherJulian, line 102: I think it would be better to log the warning but execute the command, as well. When the assertion noErrorCode does not throw the error the user may think that the command executed properly and something other caused the error.
The warning is an information for the user that the error is caused by the missing ide.

GuentherJulian (changes requested):

GuentherJulian (approved):

Issue #96, opened 2021-01-21
console runner: downloadFile


GuentherJulian:
Implement the command downloadFile in the console runner. Gets the url of the file and the path where the file should be saved as parameters.

PR #101, opened 2021-01-27
WIP: feature/consoleDownloadFile

MarcelDiessner:
#96

### DIFF
--- a/runners/console/index.ts
+++ b/runners/console/index.ts
@@ -257,12 +257,17 @@ export class Console extends Runner {
             installDir = path.join(installDir, command.parameters[1]);
             this.createFolder(installDir, false);
         }
+        var command1;
+        var command2;
         if(this.platform == ConsolePlatform.WINDOWS) {
-            this.executeCommandSync("powershell.exe \"Invoke-WebRequest -OutFile download.tar.gz '" + downloadUrl + "'\"", installDir, result);
-            this.executeCommandSync("powershell.exe tar -xvzf download.tar.gz", installDir, result);
+            command1 = "powershell.exe \"Invoke-WebRequest -OutFile download.tar.gz '" + downloadUrl + "'\"";
+            command2 = "powershell.exe tar -xvzf download.tar.gz";
         } else {
-            this.executeCommandSync("wget -c " + downloadUrl + " -O download.tar.gz | tar -xz", installDir, result);
+            command1 = "wget -c " + downloadUrl + " -O download.tar.gz";
+            command2 = "tar -xvzf download.tar.gz";
         }
+        this.executeCommandSync(command1, installDir, result);
+        this.executeCommandSync(command2, installDir, result);
         
         return result;
     }
--- a/runners/console/index.ts
+++ b/runners/console/index.ts
@@ -252,7 +252,7 @@ export class Console extends Runner {
         result.returnCode = 0;
 
         let downloadUrl = command.parameters[0];
-        let installDir = path.join(this.getWorkingDirectory(), "devonfw", "workspaces", "main", command.parameters[1]);
+        let installDir = path.join(this.getVariable(this.workspaceDirectory), command.parameters[1]);
         this.createFolder(installDir, false);
 
         if (command.parameters.length > 2) {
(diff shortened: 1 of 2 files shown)


runners/console/index.ts
```diff
@@ -215,6 +215,32 @@ export class Console extends Runner {
         return result;
     }
 
+    runDownloadFile(step: Step, command: Command): RunResult {
+        let result = new RunResult();
+        result.returnCode = 0;
+
+        let downloadUrl = command.parameters[0];
+        let installDir = path.join(this.getWorkingDirectory(), "devonfw", "workspaces", "main", command.parameters[1]);
```
GuentherJulian, line 223: Use the new path handling to get the working directory
https://github.com/devonfw-forge/tutorial-compiler/pull/100

runners/console/index.ts
```diff
@@ -393,6 +419,24 @@ export class Console extends Runner {
         }
     }
 
+    async assertDownloadFile(step: Step, command: Command, result: RunResult){
+        try {
+            let assert = new Assertions()
+            .noErrorCode(result)
+            .noException(result)
+            .directoryExits(path.join(this.getWorkingDirectory(), "devonfw", "workspaces", "main", command.parameters[1]))
+            .directoryNotEmpty(path.join(this.getWorkingDirectory(), "devonfw", "workspaces", "main", command.parameters[1]));
```
GuentherJulian, line 428: Use the new path handling to get the working directory

GuentherJulian (changes requested):

runners/console/index.ts
```diff
@@ -247,6 +247,32 @@ export class Console extends Runner {
         return result;
     }
 
+    runDownloadFile(step: Step, command: Command): RunResult {
+        let result = new RunResult();
+        result.returnCode = 0;
+
+        let downloadUrl = command.parameters[0];
+        let installDir = path.join(this.getVariable(this.workspaceDirectory), command.parameters[1]);
+        this.createFolder(installDir, false);
```
GuentherJulian, line 256: I think we can make the second parameter optional, too. If not parameter is specified the file will be stored into the workspace directory.

runners/console/index.ts
```diff
@@ -247,6 +247,32 @@ export class Console extends Runner {
...
+            if(this.platform == ConsolePlatform.WINDOWS) {
+                this.executeCommandSync("powershell.exe \"Invoke-WebRequest -OutFile " + name + " '" + downloadUrl + "'\"", installDir, result);
+            } else {
+                this.executeCommandSync("wget -c \"" + downloadUrl + "\" -O " + name, installDir, result);
+            }
+        } else {
+            if(this.platform == ConsolePlatform.WINDOWS) {
+                this.executeCommandSync("powershell.exe \"Invoke-WebRequest '" + downloadUrl + "'\"", installDir, result);
+            } else {
+                this.executeCommandSync("wget -c \"" + downloadUrl + "\"", installDir, result);
```
GuentherJulian, line 269: I think we can shorten this. For example save the command in a variable, add the output filename parameter if one is given and then pass this to executeCommandSync.

GuentherJulian (changes requested):

runners/console/index.ts
MarcelDiessner: Would that really make it shorter?

GuentherJulian: You then have only one call to executeCommandSync.

runners/console/index.ts
```diff
@@ -254,6 +254,30 @@ export class Console extends Runner {
...
+        let installDir = this.getVariable(this.workspaceDirectory);
+        if (command.parameters.length == 2) {
+            installDir = path.join(installDir, command.parameters[1]);
+            this.createFolder(installDir, false);
+        }
+        var command1;
+        var command2;
+        if(this.platform == ConsolePlatform.WINDOWS) {
+            command1 = "powershell.exe \"Invoke-WebRequest -OutFile download.tar.gz '" + downloadUrl + "'\"";
+            command2 = "powershell.exe tar -xvzf download.tar.gz";
```
GuentherJulian, line 271: The file can have any file format, so downloading it as ".tar.gz" would cause problems. I think we also need to pass the filename as a parameter, and then use that name.
Unzipping a zipped file is not part of this command. I think this should be implemented separately

GuentherJulian (changes requested):

runners/console/index.ts
```diff
@@ -254,6 +254,25 @@ export class Console extends Runner {
         return result;
     }
 
+    runDownloadFile(step: Step, command: Command): RunResult {
+        let result = new RunResult();
+        result.returnCode = 0;
+
+        let installDir = this.getVariable(this.workspaceDirectory);
```
GuentherJulian, line 261: Variable names should be descriptive. For example, downloadDir would be a better name.

runners/console/index.ts
```diff
@@ -254,6 +254,25 @@ export class Console extends Runner {
...
+        if (command.parameters.length == 3) {
+            installDir = path.join(installDir, command.parameters[2]);
+            this.createFolder(installDir, false);
+        }
+        var command1;
+        if(this.platform == ConsolePlatform.WINDOWS) {
+            command1 = "powershell.exe \"Invoke-WebRequest -OutFile " +   command.parameters[1] + " '" + command.parameters[0] + "'\"";
+        } else {
+            command1 = "wget -c " + command.parameters[0] + " -O " + command.parameters[1];
+        }
```
GuentherJulian, line 271: You can write this shorter using the ternary operator.
let command = (this.platform == ConsolePlatform.WINDOWS) ? ... : ...

GuentherJulian (changes requested):

GuentherJulian (approved):

PR #102, opened 2021-01-29
feature/consoleRunClientNg

GuentherJulian:


runners/console/index.ts
```diff
@@ -247,6 +254,19 @@ export class Console extends Runner {
         return result;
     }
 
+    runRunClientNg(step: Step, command: Command): RunResult {
+        let result = new RunResult();
+        result.returnCode = 0;
+
+        let projectDir = path.join(this.getVariable(this.workspaceDirectory), command.parameters[0]);
+        let process = this.executeDevonCommandAsync("ng serve", projectDir, result);
```
denise-khuu, line 262: I am not sure, but don't you have to add "npm install -g @angular/cli", when it's the first call of this function? And I think you have to use the useDevonCommand-if-else loop, we might use it independently from the DevonfwIDE.

denise-khuu (approved):

PR #121, opened 2021-02-15
feature/adaptTemplatesCobiGen

denise-khuu:


runners/console/index.ts
```diff
@@ -624,6 +635,20 @@ export class Console extends Runner {
         }
     }
 
+    async assertAdaptTemplatesCobiGen(step: Step, command: Command, result: RunResult) {
+        try {
+            new Assertions()
+            .noErrorCode(result)
+            .noException(result)
+            .fileExits(path.join(this.getWorkingDirectory(), "devonfw", "software", "cobigen-cli", "cg"))
+            .fileExits(path.join(this.getWorkingDirectory(), "devonfw", "software", "cobigen-cli", "class-loader-agent.jar"));
```
GuentherJulian, line 644: I think we don't need to check these two files because they should be installed with cobigen. But maybe we can check if the template folder exists.

GuentherJulian (changes requested):

runners/console/index.ts
```diff
@@ -624,6 +635,20 @@ export class Console extends Runner {
         }
     }
 
+    async assertAdaptTemplatesCobiGen(step: Step, command: Command, result: RunResult) {
+        try {
+            let homedir = os.homedir();
+            new Assertions()
+            .noErrorCode(result)
+            .noException(result)
+            .directoryExits(path.join(homedir, ".cobigen", "templates"));
```
GuentherJulian, line 644: Also use the directoryNotEmpty assertion

GuentherJulian (changes requested):

maybeec (approved):

PR #122, opened 2021-02-15
refactoring/changeCurrentDir

denise-khuu:


runners/katacoda/templates/cd.md
```diff
@@ -1,3 +1,5 @@
-<%= terminalDescr; %>
+<% if(terminalId) { %>
+Now you have to open another terminal. Click on the cd command twice and you will change to "<%= dir; %>" in terminal <%= terminalId; %> automatically. Alternatively you can click on the + next to \`IDE\`, choose the option \`Open New Terminal\` and run the cd command afterwards.
```
GuentherJulian, line 2: I think we should use the same citation marks everywhere. Here you use " for the path. One sentence later you use ` and in other templates ' is used.

GuentherJulian (changes requested):

GuentherJulian (approved):

PR #125, opened 2021-02-22
feature/npmInstall

denise-khuu:


documentation/Functions.md
```diff
@@ -159,8 +159,9 @@ path: The URL path on which is checked if the server is running
 ### npmInstall
 #### parameter 
 1. Path to the project where the dependencies from the package.json file are to be installed.
+2. (Optional) name of a package, global or local installation, array of npm arguments
```
GuentherJulian, line 162: Here we should mention that the parameter are passed via a json object.
And that each of the parameter inside the json object is optional.

runners/console/index.ts
```diff
@@ -253,17 +253,22 @@ export class Console extends Runner {
...
     runNpmInstall(step: Step, command: Command): RunResult {
         let result = new RunResult();
         result.returnCode = 0;
 
         let projectPath = path.join(this.getVariable(this.workspaceDirectory), command.parameters[0]);
+        let npmCommand = "npm install";
+        if(command.parameters.length > 1){
+            if (command.parameters[1].global) npmCommand += " -g ";
+            if (command.parameters[1].args) npmCommand += command.parameters[1].args.join(" ") + " ";
+            if (command.parameters[1].name) npmCommand += command.parameters[1].name;
```
GuentherJulian, line 267: This would not work if only the name is given, as there is not space before the name

runners/katacoda/templates/npmInstall.md
```diff
@@ -2,9 +2,11 @@
 
 <%= cdCommand; %>
 
-Install the dependencies from the package.json file.
+<% if(packageName){ %>
+Install the package <%= packageName; %> and add it to the dependencies.<% } else {%>
+Install the dependencies from the package.json file.<% } %>
 
-`<% if(useDevonCommand){ %>devon<% } %> npm install`{{execute T1}}
+`<% if(useDevonCommand){ %>devon<% } %> npm install<% if(global){ %> -g <% } %><%= packageArgs; %> <%= packageName; %>`{{execute T1}}
```
GuentherJulian, line 9: There is no space before the packageArgs argument. This will cause a problem when it is not installed globally.
And I think we should add a hint, what the -g parameter does.

GuentherJulian (changes requested):

runners/katacoda/index.ts
```diff
@@ -276,12 +276,18 @@ export class Katacoda extends Runner {
...
+        let packageTitle = (command.parameters.length > 1 && command.parameters[1].name) ? command.parameters[1].name : "the dependencies";
+        let npmCommand = {
+            "name": (command.parameters.length > 1 && command.parameters[1].name) ? command.parameters[1].name : undefined,
+            "global": (command.parameters.length > 1 && command.parameters[1].global) ? command.parameters[1].global : false, 
+            "args": (command.parameters.length > 1 && command.parameters[1].args) ? command.parameters[1].args.join(" ") : undefined
+        };
+
         this.steps.push({
-            "title": "Install the dependencies",
+            "title": "Install" + packageTitle,
```
GuentherJulian, line 287: Add a space after 'Install'

documentation/Functions.md
```diff
@@ -159,8 +159,16 @@ path: The URL path on which is checked if the server is running
 ### npmInstall
 #### parameter 
 1. Path to the project where the dependencies from the package.json file are to be installed.
-#### example 
+2. Json-object: Name of a package, global or local installation, or array of npm arguments
+* (Optional) name of a package {"name": string }
+* (Optional) global or local installation {"global" : boolean }
```
GuentherJulian, line 164: Add a hint that the default value for global is false

GuentherJulian (changes requested):

GuentherJulian (approved):

PR #126, opened 2021-02-22
Add runDownloadFile function

derochs:


runners/katacoda/index.ts
```diff
@@ -312,6 +312,15 @@ export class Katacoda extends Runner {
         return null;
     }
 
+    runDownloadFile(step: Step, command: Command): RunResult {
+        this.steps.push({
+            "title": "Download a file",
+            "text": "step" + this.stepsCount  + ".md"
+        });
+        this.renderTemplate("downloadFile.md", this.outputPathTutorial + "step" + (this.stepsCount++) + ".md", {text: step.text, textAfter: step.textAfter, downloadURL: command.parameters[0], saveDir: command.parameters[1]});
```
GuentherJulian, line 320: In the console runner the command is implemented slightly differently. The second parameter is the name of the file and the directory is the third parameter and optional. You can also see this in the function documentation.
The both runner have to use the same logic.

GuentherJulian (changes requested):

runners/katacoda/index.ts
denise-khuu: Maybe you have to add a cdCommand or a relative Path from currentDir to the logic, since it's not given, that we will execute this command from workspace. I am not sure, but it might not generate the tutorial if saveDir is missing. (the path is optional)

GuentherJulian (approved):

Issue #43, opened 2020-12-08
katacoda runner: option to add multiple functions to a step


denise-khuu:
In the asciidoc you can add only one function to a step, but in some cases the functions are designed to be just added to an existing step.

PR #128, opened 2021-02-24
feature/multiple functions per step

MarcelDiessner:
#43

### DIFF
--- a/engine/engine.ts
+++ b/engine/engine.ts
@@ -1,6 +1,7 @@
 import { Environment } from "./environment";
 import { Playbook } from "./playbook";
 import { Runner } from "./runner";
+import { RunCommand } from "./run_command";
 import { RunResult } from "./run_result";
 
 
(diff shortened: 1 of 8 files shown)


runners/console/index.ts
```diff
@@ -655,7 +654,7 @@ export class Console extends Runner {
 
         let process = child_process.spawnSync(command, { shell: true, cwd: directory, input: input, maxBuffer: Infinity, env: this.env });
         if(process.status != 0) {
-            console.log("Error executing command: " + command + " (exit code: " + process.status + ")");
+            console.log("Error executing runCommand.command: " + command + " (exit code: " + process.status + ")");
```
GuentherJulian, line 657: In strings you do not have to change it to runCommand. There are several other places in the code.

runners/katacoda/index.ts
```diff
@@ -411,4 +410,9 @@ export class Katacoda extends Runner {
         return {terminalId: this.terminalCounter, isRunning: false};
     }
 
+    private getStepsCount(runCommand: RunCommand): number {
+        let returnCount = runCommand.stepIndex == this.stepsCount - 1 ? this.stepsCount : ++this.stepsCount;
```
GuentherJulian, line 414: I tested it with your Cobigen tutorial and with the original one. Both of them do not work inside Katacoda. There are multiple steps with the same index number in the index.json. That's why Katacoda doesn't render the steps correctly.

GuentherJulian (changes requested):

runners/katacoda/index.ts
MarcelDiessner: oh i thought they should have the same stepindex because they are executed in the same step. I've done this, so that they are all written to the same file

GuentherJulian: Yes, they have to be written in the same file. But then you only need one entry for this step in the index.json. At the moment there are multiple entries with the same number.

GuentherJulian:
Resolve conflicts

engine/engine.ts
```diff
@@ -24,25 +25,36 @@ export class Engine {
...
+                if(lineIndex == 0) {
+                    runCommand.text = this.playbook.steps[stepIndex].text;
+                } 
+                if(lineIndex == (this.playbook.steps[stepIndex].lines.length - 1)){
+                    runCommand.textAfter = this.playbook.steps[stepIndex].textAfter;
+                }
+                runCommand.command = this.playbook.steps[stepIndex].lines[lineIndex];
+                runCommand.stepIndex = stepIndex;
+                runCommand.lineIndex = lineIndex;
+                runCommand.stepTitle = this.playbook.steps[stepIndex].title;
```
GuentherJulian, line 40: You can move the initialization of runCommand in an own function

runners/katacoda/index.ts
```diff
@@ -416,4 +375,25 @@ export class Katacoda extends Runner {
...
+        return returnCount;
+    }
+
+    private pushStep(runCommand: RunCommand, title: string, text: string) {
+        if (this.steps.length == this.stepsCount - 1) {
+            if (runCommand.stepTitle) {
+                this.steps.push({
+                    "title": runCommand.stepTitle,
```
GuentherJulian, line 387: When you use the ternary operator to set the title you do not need the if-else part

runners/katacoda/index.ts
```diff
@@ -416,4 +375,25 @@ export class Katacoda extends Runner {
         return {terminalId: this.terminalCounter, isRunning: false};
     }
 
+    private getStepsCount(runCommand: RunCommand): number {
+        let returnCount = runCommand.stepIndex == this.stepsCount - 1 ? this.stepsCount : ++this.stepsCount;
```
GuentherJulian, line 379: Because you use the prefix increment operator now to increment the stepCount, the false time is written into the index.json. Just remove the '-1' part of the second argument in the KatacodaTools.generateIndexJson call in the destroy method.

GuentherJulian (changes requested):

GuentherJulian (approved):

PR #133, opened 2021-03-02
WIP: feature/executeFile

denise-khuu:
I am not too sure what to assert in this feature

### DIFF
--- a/documentation/Functions.md
+++ b/documentation/Functions.md
@@ -240,3 +240,12 @@ image: Path to an image to be displayed in the katacoda step.
 adaptTemplatesCobiGen()
 
 ***
+
+### executeFile
+#### parameter
+* directory to the script file, relative to workspace 
+#### example
+executeFile("tutorial-compiler/localBuildRun.sh")
+
+***
+
--- a/documentation/Functions.md
+++ b/documentation/Functions.md
@@ -159,10 +159,10 @@ path: The URL path on which is checked if the server is running
 ### npmInstall
 #### parameter 
 1. Path to the project where the dependencies from the package.json file are to be installed.
-2. Json-object: Name of a package, global or local installation, or array of npm arguments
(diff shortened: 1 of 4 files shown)


documentation/Functions.md
```diff
@@ -240,3 +240,12 @@ image: Path to an image to be displayed in the katacoda step.
 adaptTemplatesCobiGen()
 
 ***
+
+### executeFile
+#### parameter
+* directory to the script file, relative to workspace
```
GuentherJulian, line 246: For the other commands, we use numbers to specify the parameters.
And the file does not have to be a script file. It could also be an executable file (e.g. .exe, .msi)

runners/console/index.ts
```diff
@@ -343,6 +343,13 @@ export class Console extends Runner {
         return result;
     }
 
+    runExecuteFile(step: Step, command: Command): RunResult {
+        let result = new RunResult();
+        result.returnCode = 0;
+        this.executeCommandSync("bash " + path.basename(command.parameters[0]), path.join(this.getWorkingDirectory(), path.dirname(command.parameters[0])), result);
```
GuentherJulian, line 349: You need to differentiate between Linux and Windows

runners/katacoda/templates/executeFile.md
```diff
@@ -0,0 +1,9 @@
+<%= text; %>
+
+<%= cdCommand; %>
+
+Run <%= name; %> with this bash-command.  
+
+`bash <%= name; %>`{{execute T1}}
```
GuentherJulian, line 7: It could be that the executed files start a server. Then we need to adjust the terminal handling.
We can pass this as an optional parameter (e.g. 'asynchronous: true'). And in some cases we have to pass command line arguments. For this we need also some additional paramter.

GuentherJulian (changes requested):

PR #136, opened 2021-03-02
Bug green pipeline when parser fails to parse playbook

GuentherJulian:


engine/run.ts
```diff
@@ -12,47 +12,57 @@ class Run {
...
+                let indexFile = playbooksDir + playbookDirs[index] + "/index.asciidoc";
+                if (fs.existsSync(indexFile)) {
+                    let playbook = parser.parse(indexFile);
+                    playbook.name = playbookDirs[index] + "/";
+                    playbook.path = playbooksDir + playbookDirs[index] + "/";
+                    this.playbooks.push(playbook);
+                }
             }
+        } catch(e) {
+            throw new Error("Error while parsing playbook: " + e);
```
maybeec, line 65: Why should that fix the error? You are catching an exception and rethrow it. I guess the error might be suppressed as the function is declared asynchronously.

maybeec: So you should handle the exception in the subscription of the resulting promise

maybeec (changes requested):

engine/run.ts
GuentherJulian: This is just to have a more custom error message to let the user know that the error occurred while parsing the playbook. In the run method, this error is caught and pushed to the error array. And as a result of this, the pipeline turns red when the parsing fails.

PR #140, opened 2021-03-05
vscode runner and cobigenJava command

GuentherJulian:


runners/vscode/index.ts
```diff
@@ -0,0 +1,150 @@
...
+import * as fs from 'fs';
+
+export class VsCode extends Runner {
+
+    private installedExtensions: string[] = [];
+    private env: any;
+    private vsCodeSetup: boolean = false;
+
+    init(playbook: Playbook): void {
+        this.createFolder(path.join(__dirname, "tests"), true);
```
SchettlerKoehler, line 20: Should this be moved to the WorkingDir?

runners/vscode/index.ts
```diff
@@ -0,0 +1,150 @@
...
+        this.createFolder(path.join(__dirname, "resources"), false);
+        this.env = process.env;
+    }
+
+    setupVsCode() {
```
SchettlerKoehler, line 25: Helper functions should be private and after run* and assert*

runners/vscode/index.ts
```diff
@@ -0,0 +1,150 @@
...
+    }
+
+    async assertCobiGenJava(runCommand: RunCommand, result: RunResult) {
+        new Assertions()
+        .noErrorCode(result)
+        .noException(result)
+        .fileExits(path.join(this.getWorkingDirectory(), "devonfw", "workspaces", "main", runCommand.command.parameters[0]));
+    }
+
+    private runTest(testfile: string, result: RunResult) {
```
SchettlerKoehler, line 107: Should be renamed to executeTest because it is not the method for the command Test.

runners/vscode/test-templates/runCobiGenJava.ts
```diff
@@ -0,0 +1,64 @@
...
+                        let text = await items[i].getText();
+                        if(text.indexOf("CobiGen") > -1) {
+                            await items[i].select();
+                            break;
+                        }
+                    }
+                    await sleep(50);
+                    let terminal = new TerminalView();
+                    await terminal.executeCommand("<%= cobigenTemplates %>");
+                    await sleep(30);
```
SchettlerKoehler, line 50: This looks unstable. What if the command takes longer than 30s?

GuentherJulian: Then the test continues and the assertion at the end would fail because no successful execution is reported. I don't know what else to check. I could check the output periodically and match it with the strings I use in the assertions. But what if there is output other than "exception" and "commands were executed correctly". For example, if the command fails direcly at the beginning, for example if CobiGen is not installed.

SchettlerKoehler: I would periodically check and would make the timeout e.g. 5min. So it's as fast as possible but also works when the system is slow.

SchettlerKoehler (approved):

Issue #132, opened 2021-03-02
katacoda runner: createNgProject


denise-khuu:
Implement the command createNgProject in the katacoda runner. Gets the path to project folder and the project name as parameters.

Issue #131, opened 2021-03-02
console runner: createNgProject


denise-khuu:
Implement the command createNgProject in the console runner. Gets the path to project-folder and the project name as parameters.

PR #141, opened 2021-03-08
feature/createNgProject

MarcelDiessner:
#131 & #132

### DIFF
--- a/documentation/Functions.md
+++ b/documentation/Functions.md
@@ -17,6 +17,7 @@ The following functions are already implemented:
 * downloadFile
 * nextKatacodaStep
 * adaptTemplatesCobiGen
+* createDevon4ngProject
 
 ***
 
@@ -246,5 +247,18 @@ image: Path to an image to be displayed in the katacoda step.
 * No parameters
 #### example
 adaptTemplatesCobiGen()
+
+***
+
+### createDevon4ngProject
+#### parameter 
+1. Name of the Project.
+2. (Optional) Path to where the Project should be created (relative to workspace). Folder should exist.
+#### example 
+createDevon4ngProject("exampleAngularProject")
+Will create the angular project to the current workspace with the name exampleAngularProject.
+
+createDevon4ngProject("exampleAngularProject", "projects")
+Will create the angular project to the directory projects within the current workspace.
 
 ***
--- a/documentation/Functions.md
+++ b/documentation/Functions.md
@@ -253,13 +253,13 @@ adaptTemplatesCobiGen()
 #### parameter 
 1. Name of the Project.
 2. Path to where the Project should be created (relative to workspace). Folder should exist.
-3. (Optional) Optional parameters, according to https://angular.io/cli/new .
+3. (Optional) String array of parameters, according to https://angular.io/cli/new.
 #### example 
-createDevon4ngProject("exampleAngularProject")
+createDevon4ngProject("exampleAngularProject", "")
 Will create the angular project to the current workspace with the name exampleAngularProject.
 
-createDevon4ngProject("exampleAngularProject", "projects")
-Will create the angular project to the directory projects within the current workspace.
+createDevon4ngProject("exampleAngularProject", "projects", ["--verbose"])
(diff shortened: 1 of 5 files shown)


runners/console/index.ts
```diff
@@ -347,6 +347,22 @@ export class Console extends Runner {
         return result;
     }
 
+    runCreateNgProject(runCommand: RunCommand): RunResult {
+        let result = new RunResult();
+        result.returnCode = 0;
+
+        let projectPath = (runCommand.command.parameters.length > 1) ? runCommand.command.parameters[1] : "";
+        let projectDir = path.join(this.getVariable(this.workspaceDirectory), projectPath);
+        console.log(projectDir);
```
denise-khuu, line 356: console.log Is not necessary

runners/console/index.ts
```diff
@@ -347,6 +347,22 @@ export class Console extends Runner {
         return result;
     }
 
+    runCreateNgProject(runCommand: RunCommand): RunResult {
```
GuentherJulian, line 350: I think we should name the command 'createDevon4ngProject' to keep it consistent with createDevon4jProject. We can then use the devon syntax to create the project and run 'devon ng create ...'. It is the same as 'devon ng new', but on the devon website this is the way it is used.
@denise-khuu What do you think?

GuentherJulian (changes requested):

GuentherJulian:
Don't forget to update the function documentation.

runners/console/index.ts
denise-khuu: I would change the name to 'createDevon4ngProject' but still give the option to create an angular project on it's own. We have a focus on devon4ng tutorials, but this also implies that we would have to use restoreDevonfwIde in all angular tutorials, which will take a lot of time to setup.

GuentherJulian: Do you mean to name it createDevon4ngProject but use the 'normal' way (ng new) when no ide was installed before? Or to implement a own function for this?

MarcelDiessner: For now I changed the function to createDevon4ngProject.

denise-khuu: Yes, using the normal way, if the IDE is not installed, is what I meant @GuentherJulian

runners/console/index.ts
```diff
@@ -347,6 +347,19 @@ export class Console extends Runner {
         return result;
     }
 
+    runCreateDevon4ngProject(runCommand: RunCommand): RunResult {
+        let result = new RunResult();
+        result.returnCode = 0;
+
+        let projectDir = path.join(this.getVariable(this.workspaceDirectory), runCommand.command.parameters[1]);
+        let optional = runCommand.command.parameters.length > 2 ? (" " + runCommand.command.parameters[2]) : "";
```
GuentherJulian, line 355: params or something like this would be a better name for the variable

documentation/Functions.md
```diff
@@ -240,11 +241,27 @@ image: Path to an image to be displayed in the katacoda step.
...
 #### example
 adaptTemplatesCobiGen()
 
 ***
+
+### createDevon4ngProject
+#### parameter 
+1. Name of the Project.
+2. Path to where the Project should be created (relative to workspace). Folder should exist.
+3. (Optional) Optional parameters, according to https://angular.io/cli/new .
```
GuentherJulian, line 256: We have to mention in what format the user have to pass the parameters. In you current implemention it has to be a string. I think a string array like in npmInstall would be better.

documentation/Functions.md
```diff
@@ -240,11 +241,27 @@ image: Path to an image to be displayed in the katacoda step.
...
 * No parameters
 #### example
 adaptTemplatesCobiGen()
 
 ***
+
+### createDevon4ngProject
+#### parameter 
+1. Name of the Project.
+2. Path to where the Project should be created (relative to workspace). Folder should exist.
```
GuentherJulian, line 255: Here you should mention that the path can be an empty string and that the project then will be created in the workspace directory.

GuentherJulian (changes requested):

GuentherJulian (approved):

PR #151, opened 2021-03-16
fix for vscode runner

GuentherJulian:


runners/vscode/index.ts
```diff
@@ -147,4 +162,16 @@ export class VsCode extends Runner {
             child_process.spawnSync(vsCodeBin + " --uninstall-extension " + extension, { shell: true });
         });
     }
+
+    private cleanUp(): void {
+        this.uninstallExtensions(VsCodeUtils.getVsCodeExecutable());
+        ConsoleUtils.restoreDevonDirectory();
+    }
+
+    supports(name: string, parameters: any[]): boolean {
```
SchettlerKoehler, line 171: I think it would be better to set the flag in init, because it does not rely on the fact that supports is called on all runners before everything is executed.

SchettlerKoehler (changes requested):

SchettlerKoehler (approved):

PR #154, opened 2021-03-19
ExecuteCommand

EduardKrieger:
ExecuteCommand and the bugfix are implemented


GuentherJulian:
Resolve conflicts

runners/console/index.ts
```diff
@@ -742,4 +743,36 @@ export class Console extends Runner {
...
+
+        if(runCommand.command.parameters.length > 1 && runCommand.command.parameters[1].args) 
+            exeCommandWithPath = exeCommandWithPath+ " " +runCommand.command.parameters[1].args.join(" ");
+
+        let dirPath = runCommand.command.parameters.length > 1 && runCommand.command.parameters[1].dir
+        ? path.join(this.getWorkingDirectory(), runCommand.command.parameters[1].dir)
+        : this.getWorkingDirectory();
+
+        if(runCommand.command.parameters.length > 1 && runCommand.command.parameters[1].asynchronous){
+            ConsoleUtils.executeCommandAsync(exeCommandWithPath, dirPath, result,this.env );
```
GuentherJulian, line 768: If a command is executed we have to clean everything up after the playbook was executed. So you have to pass the information to this.asyncProcesses.

GuentherJulian (changes requested):

runners/console/consoleUtils.ts
```diff
@@ -21,7 +21,7 @@ export class ConsoleUtils {
 
     static executeCommandAsync(command: string, directory: string, result: RunResult, env: any): child_process.ChildProcess {
         if(result.returnCode != 0) return;
-
+        console.log(command);
```
MarcelDiessner, line 24: Remove log

runners/console/index.ts
```diff
@@ -346,6 +346,37 @@ export class Console extends Runner {
...
+    runExecuteCommand(runCommand: RunCommand): RunResult {
+        let result = new RunResult();
+        result.returnCode = 0;
+        let exeCommand = runCommand.command.parameters[0];
+        
+        if(this.platform == ConsolePlatform.WINDOWS)
+        {
+            if(exeCommand.includes("bash "))
+            {
+                exeCommand =exeCommand.replace("bash ", ".\\");
```
GuentherJulian, line 358: Is this really needed?

runners/console/index.ts
```diff
@@ -346,6 +346,37 @@ export class Console extends Runner {
...
+            }
+        }
+
+        exeCommand = runCommand.command.parameters.length > 1 && runCommand.command.parameters[1].args
+        ? exeCommand+ " " +runCommand.command.parameters[1].args.join(" ")
```
GuentherJulian, line 363: Add an indentation when using the ternary operator

runners/console/index.ts
```diff
@@ -346,6 +346,37 @@ export class Console extends Runner {
         return result;
     }
 
+    runExecuteCommand(runCommand: RunCommand): RunResult {
```
GuentherJulian, line 349: Each run method should have an assert method where you check if the command was executed correctly.

runners/console/index.ts
```diff
@@ -410,10 +411,58 @@ export class Console extends Runner {
...
+        : exeCommand;
+
+        let dirPath = (runCommand.command.parameters.length > 1 && runCommand.command.parameters[1].dir)
+        ? path.join(this.getWorkingDirectory(), runCommand.command.parameters[1].dir)
+        : this.getWorkingDirectory();
+
+        if(runCommand.command.parameters.length > 1 && runCommand.command.parameters[1].asynchronous){
+            let process = ConsoleUtils.executeCommandAsync(exeCommand, dirPath, result,this.env);
+            if(process.pid && runCommand.command.parameters[2].port) {
+                this.asyncProcesses.push({ pid: process.pid, name: "ExecuteCommand", port: runCommand.command.parameters[2].port});
```
GuentherJulian, line 434: @denise-khuu @MarcelDiessner I think we have to think of a way to deal with this case. Otherwise we have the same problem as with the other asynchronous operations  and the buildRun script will never finish.
The port should here also be  a mandatory attribute I think. And the name is also used in the killAsyncProcesses method.  Maybe we have to remove it from there and just kill all processes on the given port

runners/katacoda/index.ts
```diff
@@ -356,6 +356,35 @@ export class Katacoda extends Runner {
         return null;
     }
 
+    runExecuteCommand(runCommand: RunCommand) : RunResult {
+        let terminal = (runCommand.command.parameters.length > 1 && runCommand.command.parameters[1].asynchronous) 
+            ? this.getTerminal('executeCommand') 
+            : undefined;
+        
+        let command = runCommand.command.parameters[0];
```
GuentherJulian, line 364: The variable 'command' is only used in the assignment of bashCommand. So can direcly use it there

GuentherJulian (changes requested):

documentation/Functions.md
```diff
@@ -22,6 +22,36 @@ The following functions are already implemented:
...
+#### example
+
+executeCommand("node" ,{"args": ["-v"]})
+Will create a command for executing node -v .
+
+executeCommand("bash someScript.sh", {"dir": "data/setup","asynchronous": "true", "args": ["--help"]})
+Will create a command to execute the script in the directory with the parameter --help and in a new Terminal.
+
+executeCommand("bash someServerScript.sh", {"asynchronous": true, "args":["-port 8080"] },{"port":8080 , "startupTime": 20, "path": "some/path/", "interval": 2})
+Starting a Server in a new Terminal with some assert information
```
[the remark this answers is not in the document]
EduardKrieger: Can I just link the Assertions.md for more Information?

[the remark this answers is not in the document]
GuentherJulian: I would like it better if you relate it to your specific example

runners/console/index.ts
EduardKrieger: Isn´t the port mandatory already? In the fix/serverIsReachable there is a check if the assert information is transferred and I check if a port is given.

MarcelDiessner: Maybe we can use a default port, which will be used if no port is given as a parameter

MarcelDiessner: @EduardKrieger Well in your implementation it is mandatory, but we need to always push the Process into the asyncProcess Array, so it can be killed later. Right now it is only pushed if a port is given.

EduardKrieger: Our documentation says you have to give a port, we can check if a port is given before we start executeCommandAsync so we can throw an Error if a port is missing.

runners/console/index.ts
```diff
@@ -410,10 +411,60 @@ export class Console extends Runner {
...
+        ? runCommand.command.parameters[0]+ " " +runCommand.command.parameters[1].args.join(" ")
+        : runCommand.command.parameters[0];
+
+        let dirPath = (runCommand.command.parameters.length > 1 && runCommand.command.parameters[1].dir)
+        ? path.join(this.getVariable(this.workspaceDirectory), runCommand.command.parameters[1].dir)
+        : this.getVariable(this.workspaceDirectory)
+
+        if(runCommand.command.parameters.length > 1 && runCommand.command.parameters[1].asynchronous){
+            let port = runCommand.command.parameters[2].port 
+            ? runCommand.command.parameters[2].port
```
denise-khuu, line 432: Why specify a default port and not just throw an error message? We should do a refactoring, if this approach is the better choice.

EduardKrieger: I was thinking about doing it but in the last Review, nobody replied to that https://github.com/devonfw-tutorials/tutorial-compiler/pull/154#discussion_r609577288 so I just used @MarcelDiessner Idea of taking a default port. There still be an Error, when the assert Methode checks for a given port and will throw an Error. The default port is just for killing the process.

GuentherJulian: The port will be different in most tutorials. It depends on the application or server you start with the command. You can't just say this is now running on port xyz.
I think the user has to specify this in their tutorials and as Denise said, throw an error if no port is specified. Because there is no way to find out what port the application is running on.

denise-khuu: In my opinion throwing an error would be better. I don't think, that the background process will be found, based on the default port and thus won't be killed in the end, but I could be wrong.

denise-khuu:
I think you need to add a second parameter to the function, so it can be executed in Linux and on Windows

runners/console/index.ts
```diff
@@ -442,10 +443,70 @@ export class Console extends Runner {
...
+        let commandIndex;
+        if(runCommand.command.parameters[0] && runCommand.command.parameters[1]){
+            commandIndex = this.platform == ConsolePlatform.LINUX ? 1 : 0;
+        }
+        else{
+            throw new Error("You have to pass a Command for Windows and Linux based OS");
+        }
+
+        let exeCommand = (runCommand.command.parameters.length > 1 && runCommand.command.parameters[2].args)
+        ? runCommand.command.parameters[commandIndex]+ " " +runCommand.command.parameters[2].args.join(" ")
```
GuentherJulian, line 462: Use indentations when using the ternary operator

runners/console/index.ts
```diff
@@ -442,10 +443,70 @@ export class Console extends Runner {
...
+
+        let dirPath = (runCommand.command.parameters.length > 1 && runCommand.command.parameters[2].dir)
+        ? path.join(this.getVariable(this.workspaceDirectory), runCommand.command.parameters[2].dir)
+        : this.getVariable(this.workspaceDirectory)
+
+        if(runCommand.command.parameters.length > 2 && runCommand.command.parameters[2].asynchronous){
+            if(runCommand.command.parameters[3].port){
+                let process = ConsoleUtils.executeCommandAsync(exeCommand, dirPath, result,this.env);
+                if(process.pid) {
+                    this.asyncProcesses.push({ pid: process.pid, name: "ExecuteCommand", port: runCommand.command.parameters[3].port});
```
GuentherJulian, line 473: I think we have to remove the name property here. In the killAsyncProcesses method we compare the name with the name of the process found on the given port and only kill them when the names match. So in this case the processes would never be killed

GuentherJulian (changes requested):

runners/katacoda/templates/executeCommand.md
```diff
@@ -0,0 +1,11 @@
+<%= text; %>
+
+<% if(!bashCommand.currentDir){ %>Change directory to the directory where we want to execute the command.
+<% if(!bashCommand.interrupt){ %> With the next command we also open a new Terminal so you have to execute it twice. <% }%>
```
denise-khuu, line 4: I would write terminal in lowercase

runners/console/index.ts
```diff
@@ -442,10 +444,70 @@ export class Console extends Runner {
...
+            .noException(result);
+            if(runCommand.command.parameters.length > 3 && runCommand.command.parameters[2].asynchronous){
+                await assert.serverIsReachable({
+                    path: runCommand.command.parameters[3].path,
+                    port: runCommand.command.parameters[3].port,
+                    interval: runCommand.command.parameters[3].interval,
+                    startupTime: runCommand.command.parameters[3].startupTime,
+                    command: this.platform == ConsolePlatform.WINDOWS 
+                    ? runCommand.command.parameters[0] 
+                    : runCommand.command.parameters[1]
```
GuentherJulian, line 500: Use indentations here

runners/console/index.ts
```diff
@@ -442,10 +444,70 @@ export class Console extends Runner {
...
-        return result;
+    runExecuteCommand(runCommand: RunCommand): RunResult {
+        let result = new RunResult();
+        result.returnCode = 0; 
+        let commandIndex;
+        if(runCommand.command.parameters[0] && runCommand.command.parameters[1]){
+            commandIndex = this.platform == ConsolePlatform.LINUX ? 1 : 0;
+        }
+        else{
+            throw new Error("You have to pass a Command for Windows and Linux based OS");
```
GuentherJulian, line 459: command in lower cases

runners/katacoda/templates/executeCommand.md
```diff
@@ -0,0 +1,11 @@
+<%= text; %>
+
+<% if(!bashCommand.currentDir){ %>Change directory to the directory where we want to execute the command.
+<% if(!bashCommand.interrupt){ %> With the next command we also open a new terminal so you have to execute it twice. <% }%>
+`cd <%= bashCommand.path; %> `{{execute T<%= bashCommand.terminalId; %> <% if (bashCommand.interrupt) { %>interrupt<% } %>}} <% } %>
+
+<% if(!bashCommand.interrupt){ %>Run <%= bashCommand.name; %> with this bash-command. <% } else { %> <%= bashCommand.name; %> is already running in terminal <%= bashCommand.terminalId; %>. Rerun the command to stop and relaunch it automatically. <% } %>
+
+`<%= bashCommand.name; %> <%= bashCommand.args; %>`{{execute T<%= bashCommand.terminalId; %> <% if (bashCommand.interrupt) { %>interrupt<% } %>}}
```
GuentherJulian, line 9: Do we need the interrupt here? Wouldn't the command already be stopped by executing the 'cd' command?

runners/console/index.ts
```diff
@@ -260,7 +262,7 @@ export class Console extends Runner {
                 content = fs.readFileSync(path.join(this.playbookPath, file), { encoding: "utf-8" });
             }
         }
-        fs.writeFileSync(filepath, content);
+        fs.writeFileSync(filepath, content, {encoding: "utf-8", flag: "w+"});
```
GuentherJulian, line 265: Is this still needed with the new fix for killing the processes?

GuentherJulian (changes requested):

runners/katacoda/templates/executeCommand.md
EduardKrieger: I have reworked the structure, now you get maximum one interrupt command

denise-khuu (dismissed):

documentation/Functions.md
```diff
@@ -114,6 +115,39 @@ will set the workspace directory to "[working directory]/devonfw/workspaces/proj
...
+#### example
+
+executeCommand("node", "node" ,{"args": ["-v"]})
+Will create a command for executing node -v .
+
+executeCommand("somePollingScript.ps1","bash somePollingScript.sh", {"dir": "data/setup","asynchronous": true, "args": ["--params 5"]})
+Will create a command to execute the script in the directory with the parameter --params 5 and in a new terminal.
+
+executeCommand("someServerScript.ps1","bash someServerScript.sh", {"asynchronous": true, "args":["-port 8080"] },{"port":8080 , "startupTime": 20, "path": "some/path/", "interval": 2})
+Starting a server in a new terminal. You have to specify the port for testing, the other parameters are optional. The startupTime can specify how long the runner will wait for a response from the server process and with interval you can set the frequenzy for the server testing. The path ist the subpath from your server that should be reached.
```
GuentherJulian, line 150: 'ist' in the 4th sentence

runners/katacoda/index.ts
```diff
@@ -399,6 +399,34 @@ export class Katacoda extends Runner {
...
+        if(runCommand.command.parameters.length > 2 && runCommand.command.parameters[2].dir){
+            filepath = runCommand.command.parameters[2].asynchronous 
+                ? path.join(this.getVariable(this.workspaceDirectory), runCommand.command.parameters[2].dir).replace(/\\/g, "/")
+                : runCommand.command.parameters[2].dir;
+            changeDir = true;
+        }
+
+        let bashCommand = {
+            "name" : runCommand.command.parameters[1],
+            "changeDir" : changeDir,
```
GuentherJulian, line 418: Can't we use the changeCurrentDir method in this case? At least you have to change the currentDir variable when changing the directory.

GuentherJulian (changes requested):

runners/katacoda/index.ts
EduardKrieger: I need also to use interrupt for some cd commands. I can add it to the changeCurrentDir method or i will just leave it like that and currentDir will be updated now.

PR #159, opened 2021-03-25
Change file

EduardKrieger:


runners/katacoda/index.ts
```diff
@@ -22,7 +22,7 @@ export class Katacoda extends Runner {
     private setupDir: string;
     private currentDir: string = path.join("/root");
     private terminalCounter: number = 1;
-    private showVsCodeIde: boolean = false;
+    private showVsCodeIde: boolean = true;
```
GuentherJulian, line 25: Why you have changed this?

runners/katacoda/index.ts
```diff
@@ -186,17 +186,35 @@ export class Katacoda extends Runner {
         let fileDir = path.join(workspaceDir, runCommand.command.parameters[0]).replace(/\\/g, "/");
         let placeholder = runCommand.command.parameters[1].placeholder ? runCommand.command.parameters[1].placeholder : "";
         let dataTarget = runCommand.command.parameters[1].placeholder ? "insert" : "replace";
-
         let content = "";
+
+        if(runCommand.command.parameters[1].lineNumber)
+        {
+            let BGscript = "#!/bin/sh \n  \n sed 'lineNumber!#PLACEHOLDER#!' textdatei -i";
```
GuentherJulian, line 193: Variables should start with lower case letters and don't use german words

runners/katacoda/index.ts
```diff
@@ -186,17 +186,35 @@ export class Katacoda extends Runner {
...
+            let number = parseInt(runCommand.command.parameters[1].lineNumber, 10);
+            let numberStr = (number == 1) ? (number.toString()+"i") : ((number-1).toString()+"a");
+            BGscript =BGscript.replace("lineNumber", numberStr);
```
GuentherJulian, line 196: It would be better to use templates for the scripts. Then you just have to pass the parameters to the renderTemplates method and don't have to replace the content here

runners/katacoda/index.ts
```diff
@@ -186,17 +186,35 @@ export class Katacoda extends Runner {
...
+            BGscript =BGscript.replace("textdatei",fileDir);
+            let FGscript = fs.readFileSync(path.join(this.getRunnerDirectory(), "templates", "scripts", "insert_foreground.sh"), {encoding: "utf-8"});
+            FGscript = FGscript.replace("filename", runCommand.command.parameters[0]);
+            fs.writeFileSync(path.join(this.getRunnerDirectory(),"templates","scripts", "insert_background.sh"),BGscript, {encoding : "utf-8"});
+            fs.writeFileSync(path.join(this.getRunnerDirectory(),"templates","scripts", "insert_foreground.sh"),FGscript, {encoding : "utf-8"});
+
+            this.renderTemplate(path.join("scripts", "insert_background.sh"), path.join(this.outputPathTutorial, "insert_background.sh"), { });
+            this.renderTemplate(path.join("scripts", "insert_foreground.sh"), path.join(this.outputPathTutorial, "insert_foreground.sh"), { });
```
GuentherJulian, line 204: If you have multiple changeFile steps, then each step would create a file with the same name. This wouldn't work. You have to differentiate between the file (e.g. add the step number)

runners/katacoda/templates/executeCommand.md
```diff
@@ -0,0 +1,7 @@
+<%= text; %>
```
GuentherJulian, line 1: This file should not be part of this pr

runners/console/index.ts
```diff
@@ -177,6 +177,26 @@ export class Console extends Runner {
                 let contentFile = fs.readFileSync(path.join(this.playbookPath, file), { encoding: "utf-8" });
                 content = content.replace(placeholder, contentFile);
             }
+        } else if(runCommand.command.parameters[1].lineNumber)
+        {
+            let lineNum = parseInt(runCommand.command.parameters[1].lineNumber,10);
```
GuentherJulian, line 182: Is the second parameter of the parseInt function needed?

documentation/Functions.md
```diff
@@ -107,17 +108,20 @@ If you want to add a string to a file:
 {"content": "[string]"}
 If you want to add different contents for the katacoda and console runner, then use the properties "fileConsole" and "fileKatacoda" or "contentConsole" and "contentKatacoda":
 {"fileConsole": "[pathToConsoleFile]", "fileKatacoda": "[pathToKatacodaFile]"}
-##### Name of the placeholder
-If you want to insert content into your code between two existing lines, take the previous line as your placeholder. Add your placeholder into the new file or string, otherwise it will be replaced entirely.
```
GuentherJulian, line 111: Why removing this part? I think this should also be part of the description.

GuentherJulian (changes requested):

denise-khuu (changes requested):

runners/console/index.ts
```diff
@@ -177,6 +177,23 @@ export class Console extends Runner {
                 let contentFile = fs.readFileSync(path.join(this.playbookPath, file), { encoding: "utf-8" });
                 content = content.replace(placeholder, contentFile);
             }
+        } else if(runCommand.command.parameters[1].lineNumber){
+            let lineNum = parseInt(runCommand.command.parameters[1].lineNumber,10);
```
GuentherJulian, line 181: Why is the second parameter needed? I think 10 is the defaut value

runners/katacoda/index.ts
```diff
@@ -9,6 +9,7 @@ import { DirUtils } from "./dirUtils";
 import * as path from 'path';
 import * as ejs from 'ejs';
 import * as fs from 'fs';
+import { timeStamp } from "console";
```
GuentherJulian, line 12: Is this needed?

denise-khuu (approved):

GuentherJulian (approved):

PR #164, opened 2021-03-29
Feature/consoleRestoreWorkspace

denise-khuu:


runners/console/index.ts
```diff
@@ -370,6 +420,23 @@ export class Console extends Runner {
        this.assertInstallDevonfwIde(runCommand, result);
     }
 
+    async assertRestoreWorkspace(runCommand: RunCommand, result: RunResult) {
+        let workspacesDir = this.getVariable(this.useDevonCommand)
+        ? path.join(this.getWorkingDirectory(), "devonfw", "workspaces")
```
GuentherJulian, line 425: Add indentations when using the ternary operator

runners/console/index.ts
```diff
@@ -86,6 +86,56 @@ export class Console extends Runner {
...
+    runRestoreWorkspace(runCommand: RunCommand): RunResult {
+        let result = new RunResult();
+        result.returnCode = 0;
+
+        let workspacesName = "workspace-" + ((runCommand.command.parameters.length > 0 && runCommand.command.parameters[0].workspace)
+            ? runCommand.command.parameters[0].workspace
+            : this.playbookName.replace("/", ""));
+
+        let workspacesDir = this.getVariable(this.useDevonCommand)
+            ? path.join(this.getWorkingDirectory(), "devonfw", "workspaces")
```
GuentherJulian, line 98: In the other methods we use 'workspaces/main' as the workspace directory. Why now only 'workspaces'?

GuentherJulian (changes requested):

runners/console/index.ts
```diff
@@ -86,6 +86,56 @@ export class Console extends Runner {
...
+            : this.getVariable(this.workspaceDirectory);
+
+        //removes all the directories and files inside workspace
+        if(this.getVariable(this.useDevonCommand))
+            ConsoleUtils.executeCommandSync("rm -r " + path.join(workspacesDir, "/*").replace(/\\/g, "/"), this.getWorkingDirectory(), result, this.env);
```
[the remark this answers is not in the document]
denise-khuu: I guess this function was meant to lay the base of a tutorial. As a starting workspace, you could say. I don't think it is helpful to the ones following this tutorial, if in the background one just load a whole workspace into the current existing project. For the case that we want to clone a repository we would still have clone repository. What do you think @SchettlerKoehler ?

runners/console/index.ts
```diff
@@ -86,6 +86,56 @@ export class Console extends Runner {
...
+
+        //copies a local repository into the workspace
+        if(runCommand.command.parameters.length > 0 && runCommand.command.parameters[0].local){
+            let forkedWorkspacesDir = path.join(this.getWorkingDirectory(),'..','..','..', workspacesName);
```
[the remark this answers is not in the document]
denise-khuu: it's starts at the 'working' folder. That's why you have to clone the local repository next to the tutorial-compiler folder. 
working/../../../workspace-name
working -> build->tutorial-compiler->folder->workspace-name

runners/console/index.ts
```diff
@@ -86,6 +86,56 @@ export class Console extends Runner {
...
+            if(fs.existsSync(forkedWorkspacesDir))
+                ConsoleUtils.executeCommandSync("cp -r " + forkedWorkspacesDir + "/. " + workspacesDir, workspacesDir, result, this.env);
```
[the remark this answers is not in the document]
denise-khuu: I just find copyFile in the fs module, but you are right I would have to differentiate those commands

runners/console/index.ts
denise-khuu: Because there could be different workspaces next to each other, that you might want to load into the tutorial. You would still work mostly in the folder levels below, but that's what I understood from Jans explanation

runners/console/index.ts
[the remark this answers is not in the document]
GuentherJulian: There is a copySync method in the fs-extra module.
https://github.com/jprichardson/node-fs-extra/blob/HEAD/docs/copy-sync.md

runners/console/index.ts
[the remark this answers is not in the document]
GuentherJulian: But [here](https://github.com/devonfw-tutorials/tutorial-compiler/pull/164/files/0259e61ae864d5b152c0ee9528a46bf5dfadacda#r603989895) you said, that there could be different workspaces next to each other. Wouldn't it delete the other workspace, when you execute the command the second time? Or do I understand it wrong?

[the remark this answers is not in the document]
denise-khuu: I believe that you might want to load multiple folders inside the "workspaces" folder, but they are all saved inside the same workspace-name repository. 
If devonfw is installed. You will have this folder structure:
working
|__devonfw
 |__workspaces 
  |__main
But we want to restore all the workspaces inside the workspace folder like this: 
|_devonfw
 |_workspaces
  |_folder1
  |_folder2 

That's why I was told to delete the workspaces folder before cloning the repository. But I could be wrong.

documentation/Structure.md
```diff
@@ -0,0 +1,90 @@
...
+    |--tutorials
+
+## Tutorials
+
+    |-- tutorials
+        |-- [playbook-name]
+            /--index.asciidoc
+
+Write your own tutorials in "./tutorials/[playbook-name]/index.asciidoc".
+You will find all the details in the tutorials [Developement](https://github.com/devonfw-forge/tutorials/wiki/Development) page.
```
EduardKrieger, line 16: maybe add playbook name should be similar like the branch in the tutorials-compiler for the CI build runs ?

runners/console/index.ts
```diff
@@ -5,7 +5,7 @@ import { Assertions } from "../../assertions";
 import { Playbook } from "../../engine/playbook";
 import { ConsolePlatform, AsyncProcess } from "./consoleInterfaces";
 import * as path from 'path';
-import * as fs from "fs";
+import * as fs from "fs-extra";
```
GuentherJulian, line 8: Are the methods from the fs module (e.g. writeFileSync or appendFileSync) also available in fs-extra module?  I thought you have to add the fs-extra module additionally.

runners/console/index.ts
```diff
@@ -93,6 +93,58 @@ export class Console extends Runner {
...
+        if(this.getVariable(this.useDevonCommand)){
+            fs.rmdirSync(path.join(workspacesDir), { recursive: true });
+            fs.mkdirSync(workspacesDir);
+        }
+
+        //copies a local repository into the workspace
+        if(runCommand.command.parameters.length > 0 && runCommand.command.parameters[0].local){
+            let forkedWorkspacesDir = path.join(this.getWorkingDirectory(),'..','..','..', workspacesName);
+            if(fs.existsSync(forkedWorkspacesDir))
+                fs.copySync(path.join(forkedWorkspacesDir, '/.',), workspacesDir)
```
GuentherJulian, line 118: Is the '/.' necessary here?
And I think there is a comma to much the end of path.join()

runners/console/index.ts
```diff
@@ -353,6 +405,13 @@ export class Console extends Runner {
         return result;
     }
 
+    runChangeWorkspace(runCommand: RunCommand): RunResult {
+        let result = new RunResult();
+        result.returnCode = 0;
+        this.setVariable(this.workspaceDirectory, path.join(this.getWorkingDirectory(), runCommand.command.parameters[0]));
```
GuentherJulian, line 411: I think it would be good to check if the given directory exist and to create it if not.
You should also add a assert method for changeWorkspace. Just to keep it equal with the other methods.

runners/console/index.ts
denise-khuu: https://www.npmjs.com/package/fs-extra
here it says, that from 'fs' is not necessary anymore

runners/console/index.ts
denise-khuu: yes, otherwise it will copy the whole folder into workspacesDir. In this case I just want to copy the content of the folder into the workspaces directory

documentation/Development.md
```diff
@@ -59,7 +61,14 @@ If you want to pass arguments to this method, you have to do this in the header
...
 ### Environment 
 flag: '-e'
-value: 'katacoda', 'console'
+value: 
+    * katacoda 
+    * test_console
+    * test_vscode
+    * wiki_console 
+    * wiki_eclipse
+    * wiki_editor
```
GuentherJulian, line 70: wiki_editor and wiki_console are no environments. These are just the runners.

documentation/Functions.md
```diff
@@ -2,6 +2,7 @@
 The following functions are already implemented:
 * installDevonfwIde
 * restoreDevonfwIde
+* restoreWorkspace
```
GuentherJulian, line 5: also add changeWorkspace

GuentherJulian (changes requested):

GuentherJulian (approved):

PR #184, opened 2021-04-01
Better error messages

SchettlerKoehler:
I changed the parser to get better error messages. See human-readable-names on https://pegjs.org/documentation

### DIFF
--- a/engine/parser.def
+++ b/engine/parser.def
@@ -3,10 +3,10 @@ start
     description
     step+
 
-headline
+headline "headline starting with ="
   = "=" _ string ___
   
-description
+description "description surrounded by ===="
   = "====" ___
     descriptionlines
   	"====" ___ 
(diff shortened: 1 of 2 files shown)


engine/parser.def
```diff
@@ -47,13 +50,16 @@ steptextafterlines
...
  
 steplines
   = stepline+ { return { "steplines": text()}; }
   
 stepline
-  = !"--" string __
+  = function __
+
+function "function call"
+  = [^\r\n(]+[(][^\r\n)]*[)] { return text(); }
```
maybeec, line 62: Shouldnt this be os agnostic line breaks?
I would recommend to add a new token for line breaks and whitespaces by default to cover all cases like this :
https://github.com/maybeec/grammars-v4/blob/master/python3/Python3.g4#L612

maybeec (changes requested):

engine/parser.def
SchettlerKoehler: As the line breaks are in [ ] they are os agnostic because this only means "not \r or \n". So if there is only a \n it will also work.

engine/parser.def
```diff
@@ -3,10 +3,10 @@ start
...
 
-headline
+headline "headline starting with ="
   = "=" _ string ___
   
-description
+description "description surrounded by ===="
   = "====" ___
     descriptionlines
   	"====" ___
```
MarcelDiessner, line 12: You could use the blockmarker here too, instead of "===="

MarcelDiessner: Also at lot of other places like !"===="

MarcelDiessner (approved):

maybeec (approved):

denise-khuu (approved):

PR #189, opened 2021-04-07
bugfix/asyncFunctions

denise-khuu:
We have to change the port to a mandatory parameter, otherwise the processes won't be killed by the killAsyncProcesses function. Without a port it won't be listed inside the asyncProcesses variable.

### DIFF
--- a/runners/console/index.ts
+++ b/runners/console/index.ts
@@ -252,17 +252,20 @@ export class Console extends Runner {
         result.returnCode = 0;
         
         let filepath = path.join(this.getVariable(this.workspaceDirectory), runCommand.command.parameters[0]);
-
-        let process = ConsoleUtils.executeCommandAsync("docker-compose up", filepath, result, this.env);
-        process.on('close', (code) => {
-            if (code !== 0) {
-                result.returnCode = code;
+        if(runCommand.command.parameters.length == 2 && runCommand.command.parameters[1].port){
+            let process = ConsoleUtils.executeCommandAsync("docker-compose up", filepath, result, this.env);
+            process.on('close', (code) => {
+                if (code !== 0) {
+                    result.returnCode = code;
+                }
+            });
+            if(process.pid) {
+                this.asyncProcesses.push({ pid: process.pid, name: "dockerCompose", port: runCommand.command.parameters[1].port });
             }
-          });
-        if(process.pid && runCommand.command.parameters.length == 2) {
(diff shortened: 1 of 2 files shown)


runners/console/index.ts
```diff
@@ -252,17 +252,20 @@ export class Console extends Runner {
...
+            });
+            if(process.pid) {
+                this.asyncProcesses.push({ pid: process.pid, name: "dockerCompose", port: runCommand.command.parameters[1].port });
             }
-          });
-        if(process.pid && runCommand.command.parameters.length == 2) {
-            this.asyncProcesses.push({ pid: process.pid, name: "dockerCompose", port: runCommand.command.parameters[1].port });
+        }else{
+            result.returnCode = 1; 
+            console.error("Missing arguments in runDockerCompose(). You have to specify a port for the server. For further information read the function documentation.")
```
MarcelDiessner, line 267: I think it should work without a given port aswell.

runners/console/index.ts
```diff
@@ -252,17 +252,20 @@ export class Console extends Runner {
...
+            process.on('close', (code) => {
+                if (code !== 0) {
+                    result.returnCode = code;
+                }
+            });
+            if(process.pid) {
+                this.asyncProcesses.push({ pid: process.pid, name: "dockerCompose", port: runCommand.command.parameters[1].port });
             }
-          });
-        if(process.pid && runCommand.command.parameters.length == 2) {
```
MarcelDiessner, line 262: I think we should check additionally if the port is given. Only checking if we got two parameters is not enough, because what if we give a second parameter object which does not include port. Then port will be null.

denise-khuu: It's already inside  a if condition, where we check if the port was given

runners/console/index.ts
denise-khuu: The bug found in the last tutorial, we talked about, wasn't actually fixed in the last PR. ExecuteCommand("mvn spring-boot:run") is running as implemented before in the background independently from the parameters that were given. If we don't specify a port, the process won't be listed inside the asyncProcesses variable and thus won't be killed at the end. Consequently, the buildRun.sh file will run endlessly, when not closed manuelly. That's why the port variable has to be mandatory

runners/console/index.ts
denise-khuu: this is the code that was overwritten

GuentherJulian (approved):

Issue #178, opened 2021-03-31
wiki runner: npmInstall


derochs:
Implement the command `npmInstall` in the wiki runners.

**Parameters:**
1. Path to project containing `package.json` of project: string
2. Json-object: Name of a package (local or global) or array of npm arguments
* (Optional) Name of package
* (Optional) Local or global installation, default is local
* (Optional) Array of npm arguments as json-object

More information [here](https://github.com/devonfw-tutorials/tutorial-compiler/wiki/Functions#npminstall)

**Wiki Console**
Install NPM dependencies from terminal including all given parameters.

**Wiki Editor**
Nothing to do

**Wiki VS Code**
Nothing to do

**Wiki Eclipse**
Nothing to do

PR #192, opened 2021-04-09
wikirunner/npmInstall

MarcelDiessner:
#178

### DIFF
--- a/runners/wikiConsole/templates/npmInstall.asciidoc
+++ b/runners/wikiConsole/templates/npmInstall.asciidoc
@@ -1,5 +1,5 @@
-npmInstall
-==========
+Installation of packages with npm install
+=========================================
 
 Prerequisites
 ~~~~~~~~~~~~~~
--- a/runners/wikiConsole/templates/npmInstall.asciidoc
+++ b/runners/wikiConsole/templates/npmInstall.asciidoc
@@ -1,0 +1,19 @@
+Install npm
+===========
+
+Prerequisites
+~~~~~~~~~~~~~~
+First make sure to have installed npm. We recommend to install the devonfwIDE which includes an npm installation. Alternatively, you can download npm https://www.npmjs.com/get-npm[here] and follow the installation instructions. 
+Then, you need to be located in the project directory where the package.json file lies.
+For this tutorial it is `<%= projectPath; %>`. You can either move there manually and open the terminal there or open the terminal and move there by executing `cd <%= projectPath; %>`.
+
+Execution
+~~~~~~~~~
+<% if(npmCommand.name){ %>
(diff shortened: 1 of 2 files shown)


runners/wikiConsole/templates/npmInstall.asciidoc
```diff
@@ -0,0 +1,19 @@
+Install npm
```
GuentherJulian, line 1: You do not install npm in this step

runners/wikiConsole/templates/npmInstall.asciidoc
```diff
@@ -0,0 +1,19 @@
...
+===========
+
+Prerequisites
+~~~~~~~~~~~~~~
+First make sure to have installed npm. We recommend to install the devonfwIDE which includes an npm installation. Alternatively, you can download npm https://www.npmjs.com/get-npm[here] and follow the installation instructions.
```
GuentherJulian, line 6: npm and nodejs have to be installed

runners/wikiConsole/templates/npmInstall.asciidoc
```diff
@@ -0,0 +1,19 @@
+Install npm
+===========
+
+Prerequisites
+~~~~~~~~~~~~~~
+First make sure to have installed npm. We recommend to install the devonfwIDE which includes an npm installation. Alternatively, you can download npm https://www.npmjs.com/get-npm[here] and follow the installation instructions. 
+Then, you need to be located in the project directory where the package.json file lies.
```
GuentherJulian, line 7: You can use the npmInstall command without project directory, too. For example if you want to install a package globally

GuentherJulian (changes requested):

runners/wikiConsole/templates/npmInstall.asciidoc
```diff
@@ -0,0 +1,19 @@
+npmInstall
```
GuentherJulian, line 1: The header should not only be the name of the command. It should describe shortly what is done

GuentherJulian (changes requested):

denise-khuu (approved):

runners/wikiConsole/templates/npmInstall.asciidoc
```diff
@@ -0,0 +1,19 @@
+Installation of packages with npm install
```
GuentherJulian, line 1: I would find it better if you write the name of the package here if a package is given and write 'Install dependencies' if the user should install the packages from the package.json file

GuentherJulian (changes requested):

GuentherJulian (approved):

PR #193, opened 2021-04-09
Feature/katacodaRestoreWorkspaces

denise-khuu:


runners/katacoda/index.ts
```diff
@@ -130,6 +130,28 @@ export class Katacoda extends Runner {
         return null;
     }
 
+    runRestoreWorkspace(runCommand: RunCommand): RunResult {
+        let workspacesName = "workspace-" + ((runCommand.command.parameters.length > 0 && runCommand.command.parameters[0].workspace)
+        ? runCommand.command.parameters[0].workspace
+        : this.playbookName.replace("/", "").replace(" ","-"));
```
GuentherJulian, line 136: Use indentions here

runners/katacoda/templates/scripts/restoreWorkspace.sh
```diff
@@ -0,0 +1,18 @@
...
+    git clone https://github.com/devonfw-tutorials/<%= workspace; %>.git <%= workspaceDir; %> 
+fi
+
+cd <%= workspaceDir; %>
+git checkout <%= branch; %> || true
+cd ~/.
+
+
+if ! <%= useDevonCommand; %>; then
+    cp -r ./workspaces/* /root && rm -R workspaces
```
GuentherJulian, line 17: Wouldn't it be better to have a workspace folder?
If not the files which are in the root of the cloned repository would be in the same directory as the setup folder in which the script files are located. The user should not come in contact with them I think.

denise-khuu: We would have to change the workspaceDir to "root/workspaces" in this case. I thought it would be kind of irritating if it's in the workingDir in console and in workspaces in Katacoda.

denise-khuu: Usually you would have a main folder as lowest directory in repository.  You would probably use changeWorkspace('main'), if you don't use the devonfw IDE. That's why, I wanted to keep it consistent with the console environment and keep it inside the workingDir.

denise-khuu: @derochs @MarcelDiessner @EduardKrieger What do you think? 

Either we change the workspaceDirectory of both runners to: 
With IDE
'[workingDirectory]/workspaces/main' 
Without IDE
'[workingDirectory]/workspaces' 

Or we copy the workspaces folder in the katacoda runner to the root folder, so it's consistent with the usual workspace directory. 

Personally I would prefere the second option, since it would make the file structure a little more intuitive. If one has worked with the other functions before, I think it would be easier this way to imagine the file system while writing the tutorials. But it's still just one folder, so it isn't such a huge change to to get used to.

GuentherJulian (approved):

MarcelDiessner (approved):

EduardKrieger (approved):

PR #199, opened 2021-04-14
feature/wikiRunnerDownloadFile

EduardKrieger:


runners/wikiConsole/templates/downloadFile.asciidoc
```diff
@@ -0,0 +1,22 @@
+Download a File
```
denise-khuu, line 1: I would find it better to use the filename here

runners/wikiConsole/templates/downloadFile.asciidoc
```diff
@@ -0,0 +1,22 @@
...
+===============
+
+Download
+~~~~~~~~
+First, you need to open a command prompt in your current Workspace. For Windows you can use Powershell and on Linux you can use the Terminal.
```
denise-khuu, line 6: *workspace

runners/wikiConsole/index.ts
```diff
@@ -29,4 +30,15 @@ export class WikiConsole extends WikiRunner {
...
         return null;
     }
+
+    runDownloadFile(runCommand: RunCommand): RunResult{
+        let url = runCommand.command.parameters[0];
+        let fileName = runCommand.command.parameters[1];
+        let dir = runCommand.command.parameters[2];
+        let directoryExists = dir 
+        ? fs.existsSync(path.join(this.getWorkingDirectory(), dir))
+        : undefined;
```
GuentherJulian, line 40: The wiki runner does not really create files or folders on the file system. So this is not needed I think

runners/wikiConsole/templates/downloadFile.asciidoc
```diff
@@ -0,0 +1,22 @@
...
+
+<%if(!directoryExists){ %>
+You can specify a folder for your download, but this folder should be existing. We create a folder with this command where we can load the files into.
+`mkdir <%=dir;%>`
+<% } %>
+
+Now we can start download the file with the wget command. You can read some Information about the usage on Linux https://manpages.ubuntu.com/manpages/hirsute/en/man1/wget.1.html[here].
+
+On Windows wget command is an alias for Invoke-WebRequest and mostly works like wget on Linux, but a detailed documentation can be found https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/invoke-webrequest?view=powershell-7.1[here].
```
GuentherJulian, line 15: I would find it better to use the Invoke-WebRequest command here, because I think this is the more natural way when working with Powershell. But you can add the information that wget is an alias and is also possible

GuentherJulian (changes requested):

runners/wikiConsole/index.ts
EduardKrieger: I want to check if the directory exists that I gonna download the file into.

runners/wikiConsole/templates/downloadFile.asciidoc
EduardKrieger: I changed it

runners/wikiConsole/index.ts
GuentherJulian: But the wiki runner only generate some description that the user can follow and execute later. So there is no need to check if the directory exists when the tutorial is executed in the compiler

EduardKrieger: I have removed it.

runners/wikiConsole/templates/downloadFile.asciidoc
```diff
@@ -0,0 +1,32 @@
+Download a File
+===============
+
+Download the file in a Terminal
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+First, you need to open a command prompt in your current Workspace. For Windows, you can use Powershell and on Linux, you can use the Terminal.
+
+Prerequisites
+~~~~~~~~~~~~~
+The Folder you want to load the file has to be existing. If the Folder doesn't exist you can create it from Terminal.
```
GuentherJulian, line 10: folder in lower case letters
...from the terminal

runners/wikiConsole/templates/downloadFile.asciidoc
```diff
@@ -0,0 +1,32 @@
...
+
+.OS
```
GuentherJulian, line 12: Is the '.OS' here needed?

GuentherJulian (changes requested):

runners/wikiConsole/templates/downloadFile.asciidoc
EduardKrieger: I find it easier to read and the structure makes it easier to skip over the part that is not required for you.

runners/wikiConsole/templates/downloadFile.asciidoc
```diff
@@ -0,0 +1,32 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+First, you need to open a command prompt in your current Workspace. For Windows, you can use Powershell and on Linux, you can use the Terminal.
+
+Prerequisites
+~~~~~~~~~~~~~
+The folder you want to load the file has to be existing. If the folder doesn't exist you can create it from terminal.
+
+.Operating system
+. Windows + 
+Execute the command to create a Folder if not already existing `if not exists <%= dir;%> mkdir <%= dir;%>`.
```
GuentherJulian, line 14: Folder here in lower case, too

GuentherJulian (changes requested):

GuentherJulian (approved):

PR #215, opened 2021-04-19
Feature/wikiRunnerCreateFile

EduardKrieger:


MarcelDiessner (approved):

MarcelDiessner (dismissed):
Fix merge conflict

runners/katacoda/index.ts
```diff
@@ -205,7 +205,7 @@ export class Katacoda extends Runner {
         }
         this.pushStep(runCommand, "Create a new file", "step" + runCommand.stepIndex + ".md");
         
-        this.renderTemplate("createFile.md", this.outputPathTutorial + "step" + runCommand.stepIndex + ".md", { text: runCommand.text, textAfter: runCommand.textAfter, filePath: filePath, fileDir: fileDir , fileName:fileName , content: content});
```
denise-khuu, line 208: I am not sure, but did you delete the outer bracket?

runners/wikiConsole/templates/createFile.asciidoc
```diff
@@ -0,0 +1,30 @@
+== Create the file <%= fileName; %>
+
+==== Prerequisites
+First, you need to open a command prompt in your current workspace. For Windows, you can use PowerShell and on Linux, you can use the Terminal.
+
+The folder where you want to create the file has to be existing. If the folder doesn’t exist you have to create it.
```
denise-khuu, line 6: I think 'has to exist' or 'You need to create the base folder, if it doesn't exist already. ', would be better

runners/katacoda/index.ts
EduardKrieger: Yes

runners/wikiConsole/templates/createFile.asciidoc
```diff
@@ -0,0 +1,30 @@
...
+
+.Operating system
+. Windows + 
+Execute the command to create a folder if not already existing `New-Item -ItemType Directory -Force -Path <%= filePath; %>`.
+. Linux + 
+Execute the command `mkdir -p <%= filePath; %>`. The `-p` flag won't throw an error if the target folder already exists and allows the creation of parent directories.
+
+=== Create <%= fileName; %> in a Terminal
```
denise-khuu, line 14: I think we need to be more consistent on how terminal is written. I've seen it in lowercase and camelcase in our templates.

runners/wikiConsole/templates/createFile.asciidoc
```diff
@@ -0,0 +1,30 @@
...
+Now you can start to create <%= fileName; %>.
+
+.Operating System
+. Windows + 
+In Powershell, we use the command *New-Item* to create a new file<% if(contentPath) { %>, and to insert some content into the created file we use the *Get-Content* command<% } %>.
+More information about the *New-Item* function and the usage can be found https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.management/new-item?view=powershell-7.1[here]<% if(contentPath) { %> and some information about the *Get-Content* function and the usage can be found https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.management/set-content?view=powershell-7.1[here]<% } %>. 
+To create a new empty <%= fileName; %> <% if(contentPath){ %>and insert some data from into it, <%= contentFile; %> <% } %>you can execute the command `New-Item  -Path "<%= filePath; %>" -Name 
+<%= fileName; %> -ItemType "file" <% if(contentPath) { %> -Value $(Get-Content -Path <%= contentPath; %>)`<% } %>. 
+
+. Linux +
```
denise-khuu, line 25: What about MacOS?

runners/wikiConsole/templates/createFile.asciidoc
EduardKrieger: I think because sometimes if you refer to the Linux Terminal is a name and in other cases you are just referring to terminals. But then its still wrong in this case.

EduardKrieger: I will use command prompt instead

runners/wikiConsole/templates/createFile.asciidoc
EduardKrieger: MacOS is added

runners/wikiConsole/templates/createFile.asciidoc
```diff
@@ -0,0 +1,30 @@
...
+In the Terminal, we start using the function *cat* to create a new file <% if(contentPath) { %>and also write the content from another file into it <% } %> 
+more information about the usage you can get https://man7.org/linux/man-pages/man1/cat.1.htm[here].<% if(contentPath) { %>
+Tho create a new empty <%= fileName; %> and write some data from <%= contentFile; %> into it, you can execute the command `cat <%= contentPath; %> > <%= filePath; %>`.
```
MarcelDiessner, line 28: *To

MarcelDiessner (changes requested):

runners/wikiVsCode/templates/createFile.asciidoc
```diff
@@ -0,0 +1,23 @@
+== Create the file <%= fileName; %>
+
+
+=== Creating <%= fileName; %> in the VSCode Editor
+
+Create <%= fileName; %> in the VSCode editor<% if(contentFile) { %> and insert some data from <%= contentFile; %> into it.
+
+First, you want to open <%= contentFile; %> to get the data we want to put into <%= fileName; %>.
+Go to the file context menu in the top left corner right next to the VS Code symbol and choose *Open File...* or use the keyboard shortcut ctrl+o.
```
denise-khuu, line 9: *VSCode

runners/wikiVsCode/templates/createFile.asciidoc
```diff
@@ -0,0 +1,23 @@
+== Create the file <%= fileName; %>
+
+
+=== Creating <%= fileName; %> in the VSCode Editor
+
+Create <%= fileName; %> in the VSCode editor<% if(contentFile) { %> and insert some data from <%= contentFile; %> into it.
+
+First, you want to open <%= contentFile; %> to get the data we want to put into <%= fileName; %>.
+Go to the file context menu in the top left corner right next to the VS Code symbol and choose *Open File...* or use the keyboard shortcut ctrl+o. 
+Based on your operating system a window with the file explorer opens.Navigate to <%= contentFile; %> and select it. Also, you can insert the path `<%= contentPath; %>` into the path bar, so you can jump into the right folder and select the file `<%= contentFile; %>`.
```
denise-khuu, line 10: opens. Navigate

runners/wikiVsCode/templates/createFile.asciidoc
denise-khuu: I think we should use either 'VSCode' or 'VS Code'

EduardKrieger: I will change it to VS Code.

runners/wikiVsCode/templates/createFile.asciidoc
```diff
@@ -0,0 +1,23 @@
...
+
+Opening a new file can be done by going to the file context menu in the top left corner right next to the VS Code symbol and selecting *New File* or use the keyboard shortcut ctrl+n and a new VS Code Editor windows will be opened with an untitled-1 filename.
+<% if(contentPath) { %> 
+Insert the previously copied text with the keyboard shortcut ctrl+v or with a right-click in the Editor window and the paste function. <% } %>
+
+The next step is to save the file by selecting *Save* or *Save as* in the file context menu or by using the keyboard shortcut ctrl+s.
+A file explorer window opens.
+You should check if you are currently in the right directory where you want to save <%= filePath; %><%= fileName;%>. 
+Changing the directory can be done by inserting the path `<%= filePath; %>` in the path bar. If the directory does not exist, create the missing folders.
+To save the file specify the name of the File. Paste `<%= fileName; %>` in the text field *File name:*.
```
denise-khuu, line 22: file

denise-khuu (approved):

runners/wikiConsole/templates/createFile.asciidoc
```diff
@@ -0,0 +1,28 @@
...
+
+=== Create <%= fileName; %> from command prompt
+First, you need to open a command prompt in your current workspace. For Windows, you can use PowerShell and on Linux, you can use the Terminal.
+
+.Operating System
+. Windows + 
+In Powershell, we use the command *New-Item* to create a new file<% if(contentPath) { %>, and to insert some content into the created file we use the *Get-Content* command<% } %>.
+More information about the *New-Item* function and the usage can be found https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.management/new-item?view=powershell-7.1[here]<% if(contentPath) { %> and some information about the *Get-Content* function and the usage can be found https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.management/set-content?view=powershell-7.1[here]<% } %>. 
+To create a new empty <%= fileName; %> <% if(contentPath){ %>and insert some data from into it, <%= contentFile; %> <% } %>you can execute the command `New-Item  -Path "<%= filePath; %>" -Name
```
GuentherJulian, line 20: Is here a word missing? `To create a new empty <%= fileName; %>`

runners/wikiConsole/templates/createFile.asciidoc
```diff
@@ -0,0 +1,28 @@
...
+
+.Operating System
+. Windows + 
+In Powershell, we use the command *New-Item* to create a new file<% if(contentPath) { %>, and to insert some content into the created file we use the *Get-Content* command<% } %>.
+More information about the *New-Item* function and the usage can be found https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.management/new-item?view=powershell-7.1[here]<% if(contentPath) { %> and some information about the *Get-Content* function and the usage can be found https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.management/set-content?view=powershell-7.1[here]<% } %>. 
+To create a new empty <%= fileName; %> <% if(contentPath){ %>and insert some data from into it, <%= contentFile; %> <% } %>you can execute the command `New-Item  -Path "<%= filePath; %>" -Name 
+<%= fileName; %> -ItemType "file" <% if(contentPath) { %> -Value $(Get-Content -Path <%= contentPath; %>)`<% } %>. 
+
+. Linux and macOS +
+In the Terminal, we start using the function *cat* to create a new file <% if(contentPath) { %>and also write the content from another file into it <% } %>
```
GuentherJulian, line 24: I think the term 'command' is more common than 'function' in the context of command lines.

runners/wikiEclipse/templates/createFile.asciidoc
```diff
@@ -0,0 +1,18 @@
+== Create the file <%= fileName; %>
+
+=== Creating <%= fileName; %> in the Eclipse IDE
+
+Create <%= fileName; %> in the Eclipse IDE<% if(contentFile) { %> and insert some data from <%= contentFile; %> into it.
+
+First, you want to open <%= contentFile; %> to get the data we want to put into <%= fileName; %>.
```
GuentherJulian, line 7: The variable `contentFile` contains the path to the file with the content you want to insert, right? But when the user goes through a wiki tutorials, he doesn`t have the file, or am I thinking wrong right now?
So he can't open the file. I think we have to get the content from the file and then paste the whole content in the template.
Same for the other templates

GuentherJulian (changes requested):

runners/wikiConsole/templates/createFile.asciidoc
```diff
@@ -0,0 +1,28 @@
...
+To create a new empty <%= fileName; %> <% if(contentPath){ %>and insert some data from into it, <%= contentFile; %> <% } %>you can execute the command `New-Item  -Path "<%= filePath; %>" -Name 
+<%= fileName; %> -ItemType "file" <% if(contentPath) { %> -Value $(Get-Content -Path <%= contentPath; %>)`<% } %>. 
+
+. Linux and macOS +
+In the Terminal, we start using the function *cat* to create a new file <% if(contentPath) { %>and also write the content from another file into it <% } %> 
+more information about the usage you can get https://man7.org/linux/man-pages/man1/cat.1.htm[here].<% if(contentPath) { %>
+To create a new empty <%= fileName; %> and write some data from <%= contentFile; %> into it, you can execute the command `cat <%= contentPath; %> > <%= filePath; %>`.
```
MarcelDiessner, line 26: Here is a word missing aswell

runners/wikiConsole/templates/createFile.asciidoc
EduardKrieger: _To create a new empty CustomerEntity.java and write some data from CustomerEntity.java into it, you can execute the command `cat <PATH>`_ Would be the created sentence and if you don't pass a content file its just: _To create a new empty CustomerEntity.java you can execute the command `cat <PATH>`_ . Do you want to have file before or after the filename ?

runners/wikiConsole/templates/createFile.asciidoc
EduardKrieger: _To create a new empty CustomerEntity.java and write some data from CustomerEntity.java into it, you can execute the command `cat <PATH>`_ Would be the created sentence and if you don't pass a content file its just: _To create a new empty CustomerEntity.java you can execute the command `cat <PATH>`_ . Do you want to have file before or after the filename ?

runners/wikiEclipse/templates/createFile.asciidoc
EduardKrieger: I can implement another case 
1. The file is in a subdirectory of the working directory and the user can get the content himself like I implemented it now 
2. The file is outside the working directory and the user has to copy the content of the file from the wiki and insert it into the file.
Or use the second case for every file, but then you have some ugly and long commands. What is your opinion?

GuentherJulian: The user has no working directory in this sense. The wiki tutorial created is made available to the user in some form and he only has this description, no other files.  So he can't just open the other file to get the content. The content of the file must be displayed within the wiki. Then he can just copy and paste it.
It's the same way as in katacoda. The content is displayed in the tutorial text on the left-hand side and can then be copied from there into the editor.

runners/wikiConsole/templates/createFile.asciidoc
GuentherJulian: See my other comment https://github.com/devonfw-tutorials/tutorial-compiler/pull/215#discussion_r635212715

runners/wikiEclipse/templates/createFile.asciidoc
denise-khuu: I think the key aspect is, that the tutorial must be understandable without having to do the tutorial actively. So all content that was changed from one file to another must be visible and comprehensible.The content file is usually not available to the user as file, that's why we need to print the content.

runners/wikiConsole/templates/createFile.asciidoc
denise-khuu: I think the file second CustomerEntity.java file is just a 'content helper' to save the content, that should be inserted into this and that position inside a file. That's why we shouldn't address it directly. 
Something like: 

First create a file like [some text]
add the following content: 

[content]

and save the new file. 

Would be better.

denise-khuu (approved):

GuentherJulian (dismissed):

MarcelDiessner (dismissed):

denise-khuu (approved):

PR #218, opened 2021-04-21
Feature/wiki runner run server java

derochs:


runners/wikiConsole/templates/runServerJava.asciidoc
```diff
@@ -0,0 +1,28 @@
+Run Java Server
+=============
```
GuentherJulian, line 2: Line too short, so it is not displayed as header

GuentherJulian (changes requested):

GuentherJulian (approved):

PR #223, opened 2021-04-23
bugfix/stepCount

### LINKED ISSUE #222: Bug: Steps are not rendered
If there are two commands in one step and the first command is a command, which does not render something (e.g. restoreDevonfwIde), the second command will not be displayed in katacoda.

```
[step]
--
restoreDevonfwIde(["java", "vscode"])
cloneRepository("", "https://github.com/devonfw/jump-the-queue.git")
--
```

In this case the 'cloneRepository' step is not shown in katacoda.

MarcelDiessner:
closes #222

### DIFF
--- a/runners/katacoda/index.ts
+++ b/runners/katacoda/index.ts
@@ -15,7 +15,7 @@ export class Katacoda extends Runner {
     private outputPathTutorial: string;
     private tempPath: string;
     private tempPathTutorial: string;
-    private stepsCount = 0;
+    private currentStepIndex = -1;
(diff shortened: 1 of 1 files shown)


GuentherJulian:
Have you tested this with the available tutorials and does it give the right output?

MarcelDiessner:
> Have you tested this with the available tutorials and does it give the right output?

Yes, i tested it and it worked as it should

GuentherJulian (approved):

PR #224, opened 2021-04-23
feature/debug mode

### LINKED ISSUE #219: console runner: debug mode
Add a debug mode which can be activated in CI pipeline with an optional parameter `--debug` to give more detailed logs and error messages, especially for asynchronous processes

MarcelDiessner:
closes #219


MarcelDiessner:
The changes in the workflow files enable us to manually start a custom run with additional settings. In this issue i introduced the --debug tag. This workflow dispatch is only possible on the default branch, so it's not possible at the moment on the main fork, but you can see the result on my fork https://github.com/MarcelDiessner/tutorial-compiler/runs/2419373959?check_suite_focus=true

.github/workflows/linuxPullRequest.yml
```diff
@@ -2,6 +2,10 @@ name: CI Linux PR
 on:
   pull_request:
     branches: [ main ]
+  workflow_dispatch:
+    inputs:
+      tags:
+        description: 'Test scenario tags'
```
GuentherJulian, line 8: I would find it better to name it 'Log level' and maybe you can provide 'debug' as the default value

.github/workflows/windowsPullRequest.yml
```diff
@@ -20,5 +24,5 @@ jobs:
       - name: run checkout.ps1
         run: .\checkout.ps1
 
-      - name: run buildRun.ps1 --user ${{ github.actor }} --branch ${{ github.event.pull_request.head.ref }}
-        run: .\buildRun.ps1 --user ${{ github.actor }} --branch ${{ github.event.pull_request.head.ref }}
+      - name: run buildRun.ps1 --user ${{ github.actor }} --branch ${{ github.event.pull_request.head.ref }} ${{ github.event.inputs.tags }}
+        run: .\buildRun.ps1 --user ${{ github.actor }} --branch ${{ github.event.pull_request.head.ref }} ${{ github.event.inputs.tags }}
```
GuentherJulian, line 28: I think we need this also on the other two workflows which run nightly and when pushing something on the main branch. Otherwise we only debug the pull request builds

engine/run.ts
```diff
@@ -17,7 +17,10 @@ class Run {
             this.parseArgs();
             this.parsePlaybooks();
             this.parseEnvironments();
-            let entries = this.filterEnv(Array.from(this.environments.entries()))
+            let entries = this.filterEnv(Array.from(this.environments.entries()));
+            if(!this.args.has("debug")) {
+                console.debug = function(){}
```
GuentherJulian, line 22: Can't we set some tag somewhere. I would find this better than override functions

GuentherJulian (changes requested):

engine/run.ts
MarcelDiessner: @SchettlerKoehler

.github/workflows/linuxPullRequest.yml
```diff
@@ -2,6 +2,11 @@ name: CI Linux PR
 on:
   pull_request:
     branches: [ main ]
+  workflow_dispatch:
```
SchettlerKoehler, line 5: I would not add this here because when you trigger this workflow by hand github.event.pull_request.head.ref is not set so the workflow will not work.

.github/workflows/windowsPullRequest.yml
```diff
@@ -2,6 +2,11 @@ name: CI Windows PR
 on:
   pull_request:
     branches: [ main ]
+  workflow_dispatch:
```
SchettlerKoehler, line 5: Same here

engine/run.ts
```diff
@@ -17,7 +17,10 @@ class Run {
             this.parseArgs();
             this.parsePlaybooks();
             this.parseEnvironments();
-            let entries = this.filterEnv(Array.from(this.environments.entries()))
+            let entries = this.filterEnv(Array.from(this.environments.entries()));
+            if(!this.args.has("debug")) {
```
SchettlerKoehler, line 21: I would move this up between parseArgs and parsePlaybooks

engine/run.ts
SchettlerKoehler: Yes, this would also be possible, but this is not the easiest solution, because you have to add a logger class that wraps the checking of this tag. To have the logging consistent you then have to change all the logging to use the new class. Currently, console is something like our logger, and overwriting functions of console to disable the output of some messages is a pattern in JS. I am not saying it is the best pattern but as the intention of this solution was to have something that can be implemented fast to debug an existing error I would go for this solution and think about if we need something better.

SchettlerKoehler (changes requested):

GuentherJulian (dismissed):

GuentherJulian (approved):

PR #246, opened 2021-05-12
WIP:feature/syntaxChecking

### LINKED ISSUE #221: New parameter for syntax checking
It would be nice to add comments in the tutorials pull request when there are syntax errors in the tutorials.

For this we need to have a new paramter we can pass to the tutorial-compiler. If the parameter is set the compiler has to store the detailed error message somewhere. In addition, we then need to modify the pipeline in the tutorials repository to read this information and comment it directly into the pull request.

MarcelDiessner:
closes #221


localBuildRun.sh
```diff
@@ -9,4 +9,4 @@ if [ $? -eq 1 ]; then
     exit 1
 fi
 
-node build/engine/run.js $*
\ No newline at end of file
+node build/engine/run.js $* --commentErrors
```
denise-khuu, line 12: I think '--commentErrors' shouldn't be hardcoded. We just want to generate an syntax Error file, when an error was found in a new pull request. That's why an extra argument was needed to use it only in the syntax check workflow and not anywhere else. You would have to add this argument inside the yml-file.

localBuildRun.ps1
```diff
@@ -5,4 +5,4 @@ Copy-Item -Force -Recurse -Path $PSScriptRoot\environments\ -Destination $PSScri
 Copy-Item -Force -Recurse -Path $PSScriptRoot\runners\ -Destination $PSScriptRoot\build
 npm test
 if(-not $?) { throw 'tests failed' } 
-node $PSScriptRoot\build\engine\run.js $args
\ No newline at end of file
+node $PSScriptRoot\build\engine\run.js $args
```
denise-khuu, line 8: If we wanted to have '--commentErrors' hardcoded into the localBuildRun.sh file. Shouldn't we add it to  localBuildRun.ps1, buildRun.sh, and BuildRun.ps1 too?

engine/syntax_checker.ts
```diff
@@ -0,0 +1,20 @@
+import * as fs from 'fs';
+import path = require('path');
+
+export class SyntaxChecker {
```
EduardKrieger, line 4: If I understand it right the SyntaxChecker just prints the syntax errors to a file, maybe you choose than a better self-describing name like SyntaxErrorLogger.

EduardKrieger (approved):

Issue #183, opened 2021-03-31
wiki runner: createDevon4ngProject


derochs:
Implement the command `createDevon4ngProject` in the wiki runners.

**Parameters:**
1. Project name
2. Path where project should be created (relative to workspace directory), folder should exist
3. (Optional) String array of parameters according to https://angular.io/cli/new

**Wiki Console**
Create new project with `devon ng create` and add optional parameters.

**Wiki Editor**
Nothing to do

**Wiki VS Code**
Nothing to do

**Wiki Eclipse**
Nothing to do

PR #248, opened 2021-05-13
#183 wikiRunner/createDevon4ngProject

derochs:


runners/wikiConsole/templates/createDevon4ngProject.asciidoc
```diff
@@ -0,0 +1,15 @@
+== Create devon4ng project
+
+<% if (cdCommand) { %>
+
+=== Prerequisites
+
+As the project should be created at `<%= cdCommand; %>`, make sure that this directory exists before proceeding with the next step. 
+
```
EduardKrieger, line 8: Maybe add ng for Prerequisites. Or can also check the installed_tools and if ng is not installed in the devonfw IDE and if missing it should be installed.

runners/wikiConsole/templates/createDevon4ngProject.asciidoc
```diff
@@ -0,0 +1,15 @@
...
+<% } %>
+
+=== Creating the Angular project
+
+Use the `ng` command provided by the `devonfw-ide` to create the project <%= projectName; %>:
```
denise-khuu, line 13: We have used devonfw IDE in all the other templates.
I think, that this function should work without devonfw too. 
At least we have implemented the option in all the other runners.

PR #250, opened 2021-05-14
feature/openFile

denise-khuu:


runners/katacoda/templates/openFile.md
```diff
@@ -0,0 +1,7 @@
+Now we want to open the file <%= fileName; %>. 
+
+You can either click on this link, here: 
+
+`<%= filePath; %>`{{open}}
+
+and it will open the file automatically or switch to the IDE and open it yourself.
```
GuentherJulian, line 7: We only use the IDE in the VS Code tutorials. In the other one the simple editor is used

GuentherJulian (changes requested):

runners/katacoda/templates/openFile.md
denise-khuu: Should we make it conditional or would 'Editor' for all cases be enough? 
We would have to change it in Create File and Change File too

denise-khuu: cd.md is also referencing the IDE I would change that too

GuentherJulian (approved):

PR #257, opened 2021-05-21
bugfix/wikiRunnerChangeFile

EduardKrieger:


denise-khuu (approved):

runners/wikiEclipse/index.ts
```diff
@@ -15,29 +16,23 @@ export class WikiEclipse extends WikiRunner {
...
-        }else{
-            contentPath = runCommand.command.parameters[1].file;
-            contentString = runCommand.command.parameters[1].content;
+        let filePath = path.join(this.getVariable(this.workspaceDirectory), runCommand.command.parameters[0].replace(fileName, ""));
+        filePath = path.relative(this.getWorkingDirectory(), filePath).replace(/\\/g, "/");
+        let fileType = this.fileTypeMap.get(fileName.substr(fileName.indexOf(".")));
+        let content = undefined;
+        if(runCommand.command.parameters[1].content || runCommand.command.parameters[1].contentConsole){
+            content = runCommand.command.parameters[1].content 
+            ? runCommand.command.parameters[1].content
```
GuentherJulian, line 26: Add indentations

runners/wikiEclipse/templates/changeFile.asciidoc
```diff
@@ -1,23 +1,24 @@
...
+Select the following text and copy it with a right-click and the copy function or the keyboard shortcut crtl+c.
+[source, <%= fileType; %>]
+----
+<%= content; %>
+----
 
+
+<% if(lineNumber) { %>Inserting the content into a specific line number can be done by a left mouse click into the line *<%= lineNumber%>* and using the keyboard shortcut ctrl+v.
+<%}else if(placeholder) {%>To replace the content with a specific Placeholder you have to locate the placeholder in <%= fileName %>. The fastest way is to search through the file with ctrl+f. A *Find/Replace* window will be opened and pasting the placeholder `<%= placeholder; %>` into the text field *find*. Use the button *Find* and the placeholder will be marked in the text and you can replace him with the new content.
+<% }else{ %>Now insert the copied text with the keyboard shortcut ctrl+v or with a right-click in the Eclipse editor window of the new file and the paste function or the keyboard shortcut ctrl+v. <% } %>
```
GuentherJulian, line 23: I am not sure if this is not a bit too detailed. I don't think we need to mention that the content can be copied with right click or ctrl +c and pasted with ctrl + v. Someone who uses Eclipse must know that 😀. 
Something like 'copy the content and insert it into the editor' would be enough.
What do you think @denise-khuu

GuentherJulian (changes requested):

runners/wikiEclipse/templates/changeFile.asciidoc
denise-khuu: I agree. I think we should assume some basic knowledge, since mostly programmers will use them anyway. They could be a little bit tedious to read, if we tell them in every tutorial how to copy and paste some text. Even I might classify them into 'too long, didn't read' tutorials, after the tenth time reading that 😄. But we should go over some tutorials anyway and check if the snippets make sense combined as a whole tutorial.

denise-khuu (approved):

denise-khuu (approved):

denise-khuu (approved):

GuentherJulian (approved):

PR #272, opened 2021-06-09
bugfix katacoda runner: fixed spelling for template variables

GuentherJulian:


runners/katacoda/index.ts
```diff
@@ -463,9 +465,13 @@ export class Katacoda extends Runner {
...
     private renderTemplate(name: string, targetPath: string, variables) {
-        let template = fs.readFileSync(path.join(this.getRunnerDirectory(),"templates", name), 'utf8');
-        let result = ejs.render(template, variables);
-        fs.writeFileSync(targetPath, result, {flag: "a"});
+        try {
+            let template = fs.readFileSync(path.join(this.getRunnerDirectory(),"templates", name), 'utf8');
+            let result = ejs.render(template, variables);
+            fs.writeFileSync(targetPath, result, {flag: "a"});
+        } catch(e) {
+            this.runResult.exceptions.push(e);
```
SchettlerKoehler, line 473: You should not need that, because the exception is caught in the engine and added to the runResult

PR #287, opened 2021-06-21
Extend Workflows

derochs:
I've added a few steps to the following Workflows: 

- CI Linux
- CI Linux PR
- CI Windows
- CI Windows PR

The addition introduces Pandoc and Asciidoctor, both tools that are required to convert AsciiDoc into Markdown syntax. The function [convertAsciidocToMarkdown](https://github.com/derochs/tutorial-compiler/blob/enhancement/convertAsciidoc/engine/converter.ts#L54) requires those tools.


.github/workflows/linuxMain.yml
```diff
@@ -18,6 +18,12 @@ jobs:
 
       - uses: actions/setup-node@v2-beta
         
+      - name: install Pandoc
+        run: sudo wget https://github.com/jgm/pandoc/releases/download/2.14.0.2/pandoc-2.14.0.2-1-amd64.deb && sudo dpkg -i pandoc-2.14.0.2-1-amd64.deb
```
GuentherJulian, line 22: Can't you just use `sudo apt-get install pandoc`. Then you don't have to change this file when a newer pandoc version is released.

GuentherJulian (approved):

PR #288, opened 2021-06-21
Enhancement/convert asciidoc

derochs:
This PR includes changes that convert AsciiDoc syntax into Markdown syntax at these locations: 

- "text" and "textAfter' placeholders in [renderTemplate](https://github.com/derochs/tutorial-compiler/blob/enhancement/convertAsciidoc/runners/katacoda/index.ts#L470) will be converted into Markdown, as they may contain AsciiDoc syntax
- in the [destroy function](https://github.com/derochs/tutorial-compiler/blob/enhancement/convertAsciidoc/runners/katacoda/index.ts#L66), the playbook description will be converted into Markdown, as well as the playbook conclusion

**Note: If this PR is merged, you'll have to install Pandoc and Asciidoctor on your local machine**

**Linux**: 
- Pandoc: `sudo wget https://github.com/jgm/pandoc/releases/download/2.14.0.2/pandoc-2.14.0.2-1-amd64.deb && sudo dpkg -i pandoc-2.14.0.2-1-amd64.deb`
- Asciidoctor: `sudo apt install asciidoctor`

**Windows**: 
- Pandoc:
  - First install chocolatey with PowerShell: `Set-ExecutionPolicy Bypass -Scope Process -Force; [System.Net.ServicePointManager]::SecurityProtocol = [System.Net.ServicePointManager]::SecurityProtocol -bor 3072; iex ((New-Object System.Net.WebClient).DownloadString('https://chocolatey.org/install.ps1'))`
  - Then install pandoc: `choco install pandoc`
- Asciidoctor: 
  - First install Ruby: `choco install ruby`
  - Then install Asciidoctor: `gem install asciidoctor`

### DIFF
--- a/engine/converter.ts
+++ b/engine/converter.ts
@@ -56,7 +56,7 @@ private platform: ConsolePlatform;
          * 1. Write asciidocString to temp.adoc file
          * 2. Use AsciiDoctor to convert AsciiDoc String to DocBook XML file
          * 3. Use Pandoc to convert DocBook XML file to Markdown File
-         * 4. Read Markdown file (eventually combine with step 2)
+         * 4. Read Markdown file (eventually combine with step 2?)
          */
 
         this.createTempFile("temp.adoc", asciidocString);
@@ -67,7 +67,9 @@ private platform: ConsolePlatform;
             this.executeCommand("pandoc -f docbook -t markdown temp.xml -o temp.md", this.getTempDir());
             markdownString = this.readTempFile("temp.md");
         }else {
-            
+            this.executeCommand("asciidoctor -b docbook temp.adoc", this.getTempDir());
+            this.executeCommand("pandoc -f docbook -t markdown temp.xml -o temp.md", this.getTempDir());
+            markdownString = this.readTempFile("temp.md");
         }
         return markdownString;
     }
--- a/engine/converter.ts
+++ b/engine/converter.ts
@@ -52,26 +52,28 @@ private platform: ConsolePlatform;
     }
 
     convertAsciidocToMarkdown(asciidocString: string): string {
-        /**
-         * 1. Write asciidocString to temp.adoc file
-         * 2. Use AsciiDoctor to convert AsciiDoc String to DocBook XML file
-         * 3. Use Pandoc to convert DocBook XML file to Markdown File
-         * 4. Read Markdown file (eventually combine with step 2?)
-         */
+        if(asciidocString !== undefined) {
+            /**
+             * 1. Write asciidocString to temp.adoc file
+             * 2. Use AsciiDoctor to convert AsciiDoc String to DocBook XML file
+             * 3. Use Pandoc to convert DocBook XML file to Markdown File
+             * 4. Read Markdown file (eventually combine with step 2?)
+             */
 
-        this.createTempFile("temp.adoc", asciidocString);
+            this.createTempFile("temp.adoc", asciidocString);
 
-        let markdownString = "";
-        if(this.platform == ConsolePlatform.WINDOWS) {
-            this.executeCommand("asciidoctor -b docbook temp.adoc", this.getTempDir());
-            this.executeCommand("pandoc -f docbook -t markdown temp.xml -o temp.md", this.getTempDir());
-            markdownString = this.readTempFile("temp.md");
-        }else {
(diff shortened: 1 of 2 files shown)


GuentherJulian:
Is there a way to check if Pandoc and Asciidoctor are installed and only do the conversion if they are. Then it is not really necessary to install the tools if you only run it locally for testing. Or maybe add a new parameter to skip the conversion.

derochs:
Makes sense, will add this feature.

engine/converter.ts
```diff
@@ -0,0 +1,100 @@
...
+             * 1. Write asciidocString to temp.adoc file
+             * 2. Use AsciiDoctor to convert AsciiDoc String to DocBook XML file
+             * 3. Use Pandoc to convert DocBook XML file to Markdown File
+             * 4. Read Markdown file (eventually combine with step 2?)
+             */
+
+            this.createTempFile("temp.adoc", asciidocString);
+
+            let markdownString = "";
+            if(this.platform == ConsolePlatform.WINDOWS) {
```
GuentherJulian, line 66: Why is this if-else-block is needed? Both block execute the same commands

engine/converter.ts
```diff
@@ -0,0 +1,100 @@
...
+                if(this.commandAvailable("asciidoctor") && this.commandAvailable("pandoc")) {
+                    this.executeCommand("asciidoctor -b docbook temp.adoc", this.getTempDir());
+                    this.executeCommand("pandoc -f docbook -t gfm temp.xml -o temp.md", this.getTempDir());
+                    markdownString = this.readTempFile("temp.md");
+                }else {
+                    markdownString = this.readTempFile("temp.adoc");
```
GuentherJulian, line 72: Couldn't you use `markdownString = asciidocString` in this case? Then you do not need to call the function

engine/converter.ts
```diff
@@ -0,0 +1,100 @@
...
+             * 2. Use AsciiDoctor to convert AsciiDoc String to DocBook XML file
+             * 3. Use Pandoc to convert DocBook XML file to Markdown File
+             * 4. Read Markdown file (eventually combine with step 2?)
+             */
+
+            this.createTempFile("temp.adoc", asciidocString);
+
+            let markdownString = "";
+            if(this.platform == ConsolePlatform.WINDOWS) {
+                if(this.commandAvailable("asciidoctor") && this.commandAvailable("pandoc")) {
```
GuentherJulian, line 67: It would be better to test if both commands are available in the constructor once. Otherwise you will always test and execute the commands when you want to convert something

GuentherJulian (changes requested):

runners/katacoda/index.ts
```diff
@@ -464,6 +468,13 @@ export class Katacoda extends Runner {
     }
 
     private renderTemplate(name: string, targetPath: string, variables) {
+        if("text" in variables) {
+            variables["text"] = this.converter.convertAsciidocToMarkdown(variables["text"]);
+        }
+        if("textAfter" in variables) {
+            variables["textAfter"] = this.converter.convertAsciidocToMarkdown(variables["textAfter"]);
+        }
```
EduardKrieger, line 476: Maybe also check for playbook.description and convert it to markdown. Now we don´t have a problem with that but somebody could also use asciidoc formating in the description in the future.

EduardKrieger (changes requested):

runners/katacoda/index.ts
derochs: If I implemented it correctly, the description conversion can be found [here](https://github.com/devonfw-tutorials/tutorial-compiler/pull/288/files#diff-720eb155dc2cef159b53febe6c14ddfdd780d241514c2f056eb71a932c4146c9R69)

engine/converter.ts
```diff
@@ -0,0 +1,100 @@
...
+        }        
+    }
+
+    commandAvailable(command: string): boolean {
+        let isAvailable = false;
+        let process = child_process.spawnSync(command, {shell: true, cwd: this.getTempDir()});
+        return process.status == 0;
+    }
+
+    executeCommand(command: string, directory: string) {
```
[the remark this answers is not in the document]
derochs: Technically yes, but the executeCommandSync() from ConsoleUtils requires parameteres that I cannot provide (e.g. RunResult, env)

EduardKrieger (approved):

PR #289, opened 2021-06-23
Assert file2

dolvie1012:
A function assertFileExist(), which gets a path relative to workspace to check if a certain file exist.


GuentherJulian:
The function actually only calls the respective assertion in the assert part. So what is the use case for this?

PR #291, opened 2021-07-01
Bugfix/kill async processes

EduardKrieger:


runners/console/index.ts
```diff
@@ -6,9 +6,9 @@ import { Playbook } from "../../engine/playbook";
 import { ConsolePlatform, AsyncProcess } from "./consoleInterfaces";
 import * as path from 'path';
 import * as fs from "fs-extra";
-import * as psList from "ps-list";
 import { ConsoleUtils } from "./consoleUtils";
 
+const { listProcess } = require("process-list");
```
GuentherJulian, line 11: You need to add the package to the package.json file, don't you?
And the brackets are not needed I think.

GuentherJulian (changes requested):

package.json
```diff
@@ -14,6 +14,7 @@
     "is-reachable": "^5.0.0",
     "mocha": "^8.2.1",
     "pegjs": "^0.10.0",
+    "process-list": "^2.0.0",
```
GuentherJulian, line 17: The other ps-list package is not needed anymore and can be removed

GuentherJulian (changes requested):

runners/console/index.ts
EduardKrieger: I have removed the brackets

package.json
EduardKrieger: I have removed the unused package

GuentherJulian (approved):

PR #308, opened 2021-07-30
update function documentation createFile

suvmanda:


documentation/Functions.md
```diff
@@ -194,9 +194,11 @@ buildJava("cobigenexample", true)
 ***
 
 ### createFile <a name="createFile"></a>
+This function create a file structure at workspace directory.
```
GuentherJulian, line 197: I would prefer something like:
`This function creates a new file with a specified content. If no content is specified, an empty file is created.`

documentation/Functions.md
```diff
@@ -194,9 +194,11 @@ buildJava("cobigenexample", true)
...
 #### parameter 
-1. Path of the file to be created (relative path to the workspace directory).
-2. (Optional) Path of the file to get the content from. Relative to the playbook directory
+There are two parameters used.
+1. 1st Parameter(cobigenexample/core/src/main/java/com/example/application/cobigenexample/customermanagement/dataaccess/api/CustomerEntity.java)-Path of the file to be created (relative path to the workspace directory).
```
GuentherJulian, line 200: I think we don't need to give the example here in the parameter list as well. We give the example later in the text, so it would be duplicated. And it makes it harder to read the text

GuentherJulian (changes requested):

GuentherJulian (approved):

PR #309, opened 2021-07-30
update function documentation runServerJava

suvmanda:


documentation/Functions.md
```diff
@@ -266,9 +266,11 @@ Repository will be cloned into a newly created subdirectory devonfw-tutorials.
 
 
 ### runServerJava <a name="runServerJava"></a>
+This function will run your java application on server.Parameter and Assertion information you need to use properly, which is mentioned below.
```
GuentherJulian, line 269: This function will run your Java application. Parameter and assertion information you need to use properly, which is mentioned below. In Katacoda the command will be executed in a new terminal.

documentation/Functions.md
```diff
@@ -266,9 +266,11 @@ Repository will be cloned into a newly created subdirectory devonfw-tutorials.
...
 #### parameter 
-1. Path to the server directory within the java project.
-2. Assertion information. Only needed for the console runner to check if the server was started properly.
+There two parameters you need to use.
```
GuentherJulian, line 271: There are...

documentation/Functions.md
```diff
@@ -266,9 +266,11 @@ Repository will be cloned into a newly created subdirectory devonfw-tutorials.
...
+1. 1st Parameter(devonfw/workspaces/main/jump-the-queue/java/jtqj/server)-Path to the server directory within the java project.
```
GuentherJulian, line 272: Java with leading capital letter

GuentherJulian (changes requested):

documentation/Functions.md
```diff
@@ -266,9 +266,17 @@ Repository will be cloned into a newly created subdirectory devonfw-tutorials.
 
 
 ### runServerJava <a name="runServerJava"></a>
+This function will run your Java application on server.Parameter and assertion information you need to use properly, which is mentioned below.In Katacoda the command will be executed in a new terminal.
```
GuentherJulian, line 269: Add spaces after the end of a sentence

documentation/Functions.md
```diff
@@ -278,6 +286,8 @@ port: Port on which the server is running
 path: The URL path on which is checked if the server is running
 
 If the tutorial should be tested on the console environment, you have to specify a port.
+
+Note: To execute this function, devonfw-ide must be installed.
```
GuentherJulian, line 290: For this command the devonfw-ide is not necessarily needed. Maven has to be installed. So if someone installs Maven before (e.g. with a setup script), then the command should also work

GuentherJulian, line 290: We should mention that the command uses the Spring Boot Maven Plugin to run the Java application. So at the moment only Spring applications are supported

GuentherJulian (changes requested):

PR #328, opened 2021-09-09
Various minor fixes to the Definition of DONE

mustafaokyay:


documentation/Definition of DONE.md
```diff
@@ -9,15 +9,16 @@
...
 * I proofread my tutorial:
      1. Are the steps available and in the correct order? 
      1. Are there any spelling mistakes? 
      1. Does everything make sense?
 * I ran through the latest update of my tutorial in Katacoda-Dev and checked if everything was working as intended BEFORE requesting a review.
 
 # **Technical**
-* The files that were used in `nextKatacodaStep `are available in `.asciidoc` files and do not contain Katacoda Syntax. `nextKatacodaStep` is almost never needed.
+* The files that were used in `nextKatacodaStep `are available in `.asciidoc` files and do not contain Katacoda Syntax.
+* The name of the function `nextKatacodaStep ` has been changed to make it clearer that there is no functional content allowed. Now it is called `displayContent `
```
GuentherJulian, line 21: Why don't you just use the name `displayContent`. I think it is a bit confusing when you first mention `nextKatacodaStep` and then `displayContent`. We should remove `nextKatacodaStep` as it is not used anymore.

mustafaokyay: Then I guess I should remove the entire sentence, as the sole purpose of mentioning nextKatacodaStep here was to remind the tutorial-makers to not use it. But it was renamed entirely so we do not need to mention it at all.

GuentherJulian: Yes, just mention `displayContent` and that no Katacoda syntax should be used inside files used in this step

PR #344, opened 2022-03-07
Fix the issue that images cause problems with setup scripts

GuentherJulian:
Adresses https://github.com/devonfw-tutorials/tutorials/issues/204

### DIFF
--- a/runners/katacoda/katacodaAssetManager.ts
+++ b/runners/katacoda/katacodaAssetManager.ts
@@ -11,7 +11,7 @@ export class KatacodaAssetManager {
         this.assetDirectory = assetDir;
     }
 
-    registerFile(filepathSource: string, filepathTarget: string, katacodaDirectory: string, copyFile: boolean, copyIntoKatacodaEnvironment: boolean) {
+    registerFile(filepathSource: string, filepathTarget: string, katacodaDirectory: string, copyFile: boolean) {
         this.assetData.push({
             sourcePath: filepathSource,
             targetPath: filepathTarget,
@@ -19,7 +19,7 @@ export class KatacodaAssetManager {
             copyFile: copyFile
         });
 
-        if(copyIntoKatacodaEnvironment) {
+        if(filepathTarget && katacodaDirectory) {
             this.katacodaAssets.push({
                 file: filepathTarget.replace(/\\/g, "/"),
                 target: katacodaDirectory.replace(/\\/g, "/")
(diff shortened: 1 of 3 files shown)


runners/katacoda/katacodaAssetManager.ts
```diff
@@ -11,27 +11,29 @@ export class KatacodaAssetManager {
...
             targetPath: filepathTarget,
             katacodaDirectory: katacodaDirectory,
             copyFile: copyFile
         });
 
-        this.katacodaAssets.push({
-            file: filepathTarget.replace(/\\/g, "/"),
-            target: katacodaDirectory.replace(/\\/g, "/")
-        })
+        if(copyIntoKatacodaEnvironment) {
```
maybeec: Shouldn't this be simply `if(filpathTarget && katacodaDirectory) `?
You don't need an extra boolean for that

maybeec (changes requested):

runners/katacoda/katacodaAssetManager.ts
GuentherJulian: Yeah you're right, thats easier. Changed it

maybeec (approved):

maybeec:
@GuentherJulian do you know what's up with the build?

GuentherJulian:
> @GuentherJulian do you know what's up with the build?

@maybeec The OpenApi tutorials fail on the windows build because the Java version does not match. Windows Server 2022 has Java 8 as default version in GitHub, but Java 11 is required for this project. The Ubuntu environment already uses Java 11.
In this tutorials we do not use the devonfw-ide, so that the version of the OS is used. Created an issue for this, to use devonfw-ide instead https://github.com/devonfw-tutorials/tutorials/issues/208

maybeec:
couldn't we then simply fix it with https://github.com/actions/setup-java ?

GuentherJulian:
> couldn't we then simply fix it with https://github.com/actions/setup-java ?

That would also be a solution. My thought was that the tutorials ensure to have the correct version on their own. Independent of the environment. Maybe in the future we will have different tutorials that need different Java versions. Then we might have the same problem later.

maybeec:
OK got it. So normally the tutorial should state which devonfw ide is setup but there is no choice to impact the Java version, isn't it? I feel even Java 11 is already defaulted in the devonfw ide.
Is this tutorial not running an devonfw ide or is it an old version which is restored here or something else?
I am trying to unterstand how we get control on it. Currently I thought we already have control by the chosen devonfw ide configuration

GuentherJulian:
> OK got it. So normally the tutorial should state which devonfw ide is setup but there is no choice to impact the Java version, isn't it? I feel even Java 11 is already defaulted in the devonfw ide. Is this tutorial not running an devonfw ide or is it an old version which is restored here or something else? I am trying to unterstand how we get control on it. Currently I thought we already have control by the chosen devonfw ide configuration

Yes, Java 11 is currently the default in the devonfw-ide. We simply do not install and use the devonfw-ide in these tutorials. So in the Windows workflow, the Java version of the OS is used, which is Java 8.
At the moment we only have control on the version of the ide, but not (for example) the Java version that is installed within the ide. That could be a feature for the future. But in this case, simply installing the default devonfw-ide would fix the problem, I think.

